PR #1, opened 2024-02-29
Add a model to the chat panel

brichet:
This PR adds a model to the chat, which must be used to implement a chat panel.

It also provides a websocket handler extending this model, as one possible implementation of the model.

The idea of this change is to be able to create a CRDT model (based on this model) in jupyter_collaboration.


src/model.ts
```diff
@@ -0,0 +1,191 @@
...
+  get messageDeleted(): ISignal<IChatModel, ChatService.IMessage>;
+
+  /**
+   * Send a message, to be defined depending on the chosen technology.
+   * Default to no-op.
+   *
+   * @param message - the message to send.
+   * @returns whether the message has been sent or not, or nothing if not needed.
+   */
+  sendMessage(
```
martinRenou, line 34: Just a naming question. I'm wondering if `addMessage` would be better.

In my mind, in this place we're changing the chat model to add a new message into it. Whether it is "sent" is entirely to the chatmodel implementation detail.

```suggestion
  addMessage(
```

martinRenou (commented):
Thanks for doing this @brichet !

This looks all good to me. I only have a small comment/suggestion/question concerning the method namings in the base ChatModel

src/model.ts
martinRenou: When looking into implementing a CRDT-based chat, we'd add a message to the data structure. The sending of the changes would not be handled by the chat implementation.

martinRenou: So I wouldn't make the base ChatModel depend on `ChatService`, it seems to me `ChatService` would be an implementation detail of the websocket chat implementation?

brichet: > it seems to me ChatService would be an implementation detail of the websocket chat implementation?

It is true for part of it, but some of the types/interfaces are more generics and used by the view.
I will split it, to import only the required types.

Issue #34, opened 2024-05-16
Set up the releaser workflow


brichet:
Task list to set up the releaser workflow:
- [x] Add the `jupyterlab releaser bot` Github App (done by default)
- [x] Create a "release" environment
- [x] Add `APP_ID` variable, `NPM_TOKEN` and `APP_PRIVATE_KEY` secrets to that environment (coming from the Github app)
- [x] Configure the ruleset on branch and tag to allow the Github app to bypass it

cc @jtpio

brichet:
thanks @jtpio 
I can't see the `NPM_TOKEN` in the secrets. Is it not required too?

jtpio:
It's in the Actions secret:

![image](https://github.com/jupyterlab/jupyter-chat/assets/591645/3504a479-7d46-4cdc-ac57-f9a26b8209cd)

jtpio:
Ah, it's probably equivalent. For reference I looked at the `jupyter-collaboration` repo since it was recently set up to use the GitHub for releasing. Both `APP_PRIVATE_KEY` and `NPM_TOKEN` are accessed via `secrets.`:

https://github.com/jupyterlab/jupyter-collaboration/blob/adb51b0efd22e63c0110c08dacfc61aebb8ef559/.github/workflows/publish-release.yml#L28

https://github.com/jupyterlab/jupyter-collaboration/blob/adb51b0efd22e63c0110c08dacfc61aebb8ef559/.github/workflows/publish-release.yml#L42

jtpio:
Also it looks like we still to update the releaser workflows to use the GitHub app, as there is still a mention of `ADMIN_GITHUB_TOKEN`:

https://github.com/jupyterlab/jupyter-chat/blob/4b34bf41bcd47ff1a267a9bf85f6b884b389e2ba/.github/workflows/publish-release.yml#L29

brichet:
Yes, I saw that too, I'll open a PR with an update

brichet:
The releaser workflow failed to publish assert https://github.com/jupyterlab/jupyter-chat/actions/runs/9118046663/job/25070279419#step:5:157.
It seems to be related to a PyPI token.

jtpio:
Could it be because there is a `pyproject.toml` file, but no Python package to publish? https://github.com/jupyterlab/jupyter-chat/blob/main/pyproject.toml

Do we need this file, or could the releaser config be moved to the `package.json` and this file be deleted?

brichet:
[...]

PR #35, opened 2024-05-16
Set the release workflows to use the jupyterlab releaser bot

brichet:
Follow up #34, to use the Github app in the release workflows.

cc @jtpio


jtpio (approved):
Thanks!

PR #40, opened 2024-05-20
Stack consecutive messages from same user

### LINKED ISSUE #33: Stack messages in the UI
### Problem

When a user send several messages in a row, the username is repeated for each message.

### Proposed Solution

The messages could be stacked to avoid repeated useless information.

### Additional context

It could be `true` by default but configurable using the config of the model.

brichet:
Fix #33 
Stack the consecutive messages from the same user, to avoid duplicating avatar and user name on each message.

It adds a config entry in settings, whether to enable or not the stacked view. The default is `true`.

| NOT STACKED | STACKED |
| -------------------- | -------------- |
| not-stacked-messages-linux | stacked-messages-linux |
---------------------------------------


martinRenou (approved):
Great!

Issue #32, opened 2024-05-16
Add notifications in collaborative chat


brichet:
### Problem

There is currently no notification on new message or new chat.

### Proposed Solution

Add notifications when a chat is created, or a new message received.

It could use the notification system of Jupyterlab, and add a `*` in the tab title.

krassowski:
It may be a good idea to only show notifications if user is in a different widget

PR #43, opened 2024-05-28
Notifications for unread messages

brichet:
Fixes https://github.com/jupyterlab/jupyter-chat/issues/32

This PR adds some features (and tests) to handle unread messages:
- navigation button
- notifications (with setting to disable it)
- symbol in tab label

Peek 2024-05-31 11-38

### Code change
- includes unread messages and current messages in viewport in the model
- the model can reveive a `CommandRegistry` to trigger the notifications
- add navigation buttons
- add a new settings to disable notifications

### TODO (in follow up PRs I think):
- be able to  listen to a chat not open (this mean having a chat model opened without widget)
- save the status to avoid marking previous message as unread (@trungleduc suggested using `localStorage`)
- add button to mark a chat as unread
- add a link to the chat in notification

cc. @trungleduc @krassowski who have shown interest in this feature


martinRenou:
> be able to listen to a chat not open (this mean having a chat model opened without widget)

As discussed privately, another approach could be to have a server extension that listens for file changes on the file system (using something like https://github.com/gorakhargosh/watchdog) as a server extension, that notifies the front-end when the chat has been updated (so there may be a new message).

@davidbrochart I recall you were looking for a tool like watchdog at some point, was it related to collaboration work?

What do you think about this idea? I feel like it would be better than opening rooms for all chats on the working directory just to get notifications enabled.

packages/jupyterlab-collaborative-chat/src/widget.tsx
```diff
@@ -283,6 +309,26 @@ class ChatSection extends PanelWithToolbar {
...
+  /**
+   * Change the title when messages are unread.
+   *
+   * TODO: fix it upstream in @jupyterlab/ui-components.
+   * Updating the title create a new Title widget, but does not attach again the
+   * toolbar. The toolbar is attached only when the title widget is attached the first
+   * time.
+   */
+  private _unreadChanged = (_: IChatModel, unread: number[]) => {
+    // this.title.label = `${unread.length ? '* ' : ''}${this._name}`;
```
martinRenou, line 329: Should we remove this, as you implemented it in CSS?

brichet: Thanks for reviewing it @martinRenou.

The CSS you saw is for the main area tab.
This part is for the sidebar title, related to https://github.com/jupyterlab/jupyterlab/pull/16390.

As soon as the PR above is released, we'll be able change the minimum version of jupyterlab and uncomment this line.

brichet:
> As discussed privately, another approach could be to have a server extension that listens for file changes on the file system (using something like https://github.com/gorakhargosh/watchdog) as a server extension, that notifies the front-end when the chat has been updated (so there may be a new message).

Thanks @martin, I agree this is probably a better approach to avoid overloading the client side. And we could still listen on selected chat on client side. The only drawback I see is that it will trigger every changes in the file, like users or some metadata that could be added later (I'm thinking about AI settings for instance).

Maybe the server extension could listen to changes on the existing shared sessions server side (instead of listening for any change in all the chat files) ?
If there is no shared session for a file, this means there is no client on this chat, we can probably assume that it should not change (except for manual change on the file, which should not happen I think).
That way we could trigger a notification only when a message is added or updated.

PR #48, opened 2024-06-08
Depend on `pydantic` directly instead of `langchain`

jtpio:
Looks like `langchain` is currently used as a proxy for importing `pydantic_v1`.

So we can probably depend on `pydantic` directly and update the imports, to avoid having to depend on `langchain`.

### DIFF
--- a/packages/jupyterlab-ws-chat/pyproject.toml
+++ b/packages/jupyterlab-ws-chat/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
     "jupyterlab~=4.0",
     "traitlets>=5.0",
     "deepmerge>=1.0",
-    "langchain"
+    "pydantic>=2"
 ]
 dynamic = ["version", "description", "authors", "urls", "keywords"]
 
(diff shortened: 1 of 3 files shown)


packages/jupyterlab-ws-chat/pyproject.toml
```diff
@@ -30,7 +30,7 @@ dependencies = [
     "jupyterlab~=4.0",
     "traitlets>=5.0",
     "deepmerge>=1.0",
-    "langchain"
+    "pydantic>=2"
```
jtpio, line 33: Requiring `>=2` as per: https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features

> The Pydantic V2 package also continues to provide access to the Pydantic V1 API by importing through pydantic.v1. For example, you can use the BaseModel class from Pydantic V1 instead of the Pydantic V2 pydantic.BaseModel class:

brichet (approved):
Thanks @jtpio

PR #55, opened 2024-06-12
Use the Chat.IOptions as option for the Widget and Sidebar

### LINKED ISSUE #50: Consider making the list of params for `buildChatSidebar` an object
Currently parameters must be passed as regular parameters to the `buildChatSidebar` helper function:

https://github.com/jupyterlab/jupyter-chat/blob/f3ba6e06ea09740303a9c9ceeeea494e5fcc63f7/packages/jupyter-chat/src/widgets/chat-sidebar.tsx#L14-L18

Maybe it could be useful to provide an `options` object instead, so it's easier to grow the list of parameters in the long run?

Something that can be called as follows by user code:

```ts
buildChatSidebar({ chatModel, themeManager });
```

brichet:
Fix #50

### DIFF
--- a/packages/jupyter-chat/src/widgets/chat-widget.tsx
+++ b/packages/jupyter-chat/src/widgets/chat-widget.tsx
@@ -3,8 +3,7 @@
  * Distributed under the terms of the Modified BSD License.
  */
 
-import { IThemeManager, ReactWidget } from '@jupyterlab/apputils';
-import { IRenderMimeRegistry } from '@jupyterlab/rendermime';
+import { ReactWidget } from '@jupyterlab/apputils';
 import React from 'react';
 
 import { Chat } from '../components/chat';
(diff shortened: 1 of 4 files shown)


packages/jupyter-chat/src/widgets/chat-widget.tsx
```diff
@@ -19,33 +18,21 @@ export class ChatWidget extends ReactWidget {
...
-    return (
-      <Chat
-        model={this._model}
-        themeManager={this._themeManager}
-        rmRegistry={this._rmRegistry}
-      />
-    );
+    // The model need to be passed, otherwise it is undefined in the widget in
+    // the case of collaborative document.
+    return <Chat {...this._chatOptions} model={this._chatOptions.model} />;
```
brichet, line 34: The `model` need to be passed to the widget, the `chatOptions` is not enough in the case of collaborative document.
Probably because the model is a getter with collaborative document:

https://github.com/jupyterlab/jupyterlab/blob/130089eddb330a44f7f9d496b1e64eddcb497d6b/packages/docregistry/src/context.ts#L140-L142

Issue #51, opened 2024-06-10
Add autocomplete feature


brichet:
As suggested by @jtpio, the [autocomplete feature of jupyter-ai](https://github.com/jupyterlab/jupyter-ai/pull/810) could be ported to jupyter-chat.

### Problem

There is no auto-completion in jupyter-chat and no way setup one for a specific chat.

### Proposed Solution

Include a "plugin system" like to allow customize auto-completion in the chat input.

PR #57, opened 2024-06-14
Autocomplete commands

brichet:
This PR adds autocomplete commands to the chat component, using a registry.
It is strongly inspired by the PR made by @dlqqq in `jupyter-ai` https://github.com/jupyterlab/jupyter-ai/pull/810

Fixes https://github.com/jupyterlab/jupyter-chat/issues/51

If some autocomplete commands are provided, a list of commands will open if the first character match a specific string.

The logic is as following:
- an extension like `jupyterlab-collaborativ-chat` provides a registry for autocomplete commands, using a token.
- an extension relying on `jupyterlab-collaborativ-chat` (like possibly `jupyter-ai`) will be able to register some autocomplete commands for its own use.

The 2 extensions (collaborative and websocket chat) provide the token in a new plugin.

cc @jtpio

### DIFF
--- a/packages/jupyter-chat/src/registry.ts
+++ b/packages/jupyter-chat/src/registry.ts
@@ -117,6 +117,13 @@ export class AutocompletionRegistry {
     return this._autocompletions.delete(name);
   }
 
+  /**
+   * Remove all registered autocompletions.
+   */
+  removeAll(): void {
+    this._autocompletions.clear();
+  }
+
   private _default: string | null = null;
   private _autocompletions = new Map<string, IAutocompletionCommandsProps>();
 }
--- a/packages/jupyter-chat/src/registry.ts
+++ b/packages/jupyter-chat/src/registry.ts
@@ -24,7 +24,7 @@ export interface IAutocompletionRegistry {
   /**
    * Get the default autocompletion.
    */
-  getDefault(): IAutocompletionCommandsProps | undefined;
+  getDefaultCompletion(): IAutocompletionCommandsProps | undefined;
   /**
    * Return a registered autocomplete props.
    *
@@ -51,7 +51,7 @@ export interface IAutocompletionRegistry {
 /**
  * A registry to provide autocompleters.
  */
-export class AutocompletionRegistry {
+export class AutocompletionRegistry implements IAutocompletionRegistry {
   /**
    * Getter and setter for the default autocompletion name.
    */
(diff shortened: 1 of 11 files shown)


jtpio:
Thanks @brichet for working on this!

I'll have a closer look very soon.

packages/jupyter-chat/src/registry.ts
```diff
@@ -0,0 +1,129 @@
...
+ */
+export interface IAutocompletionRegistry {
+  /**
+   * The default autocompletion name.
+   */
+  default: string | null;
+  /**
+   * Get the default autocompletion.
+   */
+  getDefault(): IAutocompletionCommandsProps | undefined;
```
jtpio, line 27: These two properties have a similar name and it could be confusing to know what they refer to.

Should they be renamed to something else, for example `defaultName` and `defaultCompletion`?

brichet: Thanks @jtpio, I renamed the `getDefault()` to `getDefaultCompletion()`.

brichet:
BTW @jtpio, if you want to test it easily you can apply this patch, it will register a completion object in both extensions.
The completion options should be displayed if you type `:` in the input.

[autocompletion-test.txt](https://github.com/user-attachments/files/16010381/autocompletion-test.txt)

jtpio:
Thanks!

The integration tests seem to be failing consistently with this change, or was it also happening with other changes?

jtpio:
Thanks @brichet !

> apply this patch, it will register a completion object in both extensions. The completion options should be displayed if you type : in the input.

Would it be possible to regenerate the diff, as it might now be outdated?

```
$ git apply autocomplete-test.txt
error: corrupt patch at line 256
```

jtpio:
Thanks, I just tried again in a clean state and it worked :+1:

It could have been because my local state was dirty.

packages/jupyterlab-collaborative-chat/src/index.ts
```diff
@@ -42,11 +47,25 @@ import { YChat } from './ychat';
 const FACTORY = 'Chat';
 
 const pluginIds = {
+  acRegistry: 'jupyterlab-collaborative-chat:autocompletionRegistry',
```
jtpio, line 50: Maybe we could use a full name here so it's more explicit what it refers to?

```suggestion
  autocompletionRegistry: 'jupyterlab-collaborative-chat:autocompletionRegistry',
```

jtpio:
Nice, it seems to be working well when testing locally:

image

I left a minor comment above, but otherwise we can likely move forward and iterate on the API if needed, for example when integrating with Jupyter AI.

jtpio (approved):
Thanks!

jtpio:
@brichet since some exported interfaces have changed, maybe this should be released in a `0.2.0`?

brichet:
Thanks again for the review @jtpio

PR #65, opened 2024-07-09
Updated integration tests workflow

krassowski:
Updated integration tests workflow


brichet:
Thanks @krassowski

Is there any issue or context for these changes ?

krassowski:
There will be, this is the thing I mentioned on the last lab call in the unrecorded portion.

brichet:
OK thanks, I missed this call.

PR #74, opened 2024-08-14
Fix ReadTheDocs

jtpio:
Looking into fixing the failing builds on ReadTheDocs


jtpio:
The preview is now enabled and the docs seem to be building fine: 

image

PR #84, opened 2024-09-19
Fix the update snapshot workflow

brichet:
Comment the snapshot update on jupyterlab-ws-chat


brichet:
please update snapshots

brichet:
please update snapshots

PR #88, opened 2024-10-11
Improve dev script

### LINKED ISSUE #64: Improve the `dev-install` script
The top-level `package.json` has a `dev-install` script here:

https://github.com/jupyterlab/jupyter-chat/blob/025efd158ee42a9562d9f496e340feaefc90f88a/package.json#L36

However the underlying `packages` don't seem to be exposing such `script` in their own `package.json`.

It could be useful to add such script in the different `packages` to speed up the dev install, which would for example run the `jupyter labextension develop . --overwrite` command for the lab extensions.

brichet:
Fixes #64 by creating a `dev-install` command(s) in *package.json* files.

These command install the python package in dev mode, and the labextension in develop.

The PR also rename *install.sh* script to *dev-install.sh* for consistency.

### Note

I intentionally not added the `dev-uninstall` command, which leads to unstable installation, the symlinks not being removed.


jtpio (approved):
Thanks!

Maybe a follow-up improvement could be to try support developing on windows? To be more welcoming to new contributors who may be using Windows.

Not sure the bash script would run as is on Windows at the moment.

jtpio:
> Maybe a follow-up improvement could be to try support developing on windows? To be more welcoming to new contributors who may be using Windows.

For now one can use the npm scripts directly so maybe it's enough.

brichet:
> Not sure the bash script would run as is on Windows at the moment.

Right, maybe using a python script instead of bash would do it.

Issue #61, opened 2024-07-02
Path configuration for .chat files


mvallet91:
### Problem

The `.chat` files are created in the home directory of JupyterLab. It would be nice to be able to specify the path for chat files so users can put them into a `chats` folder, or admins (for instance using JupyterHub) can store them in a different path to minimize confusion for novice users.

### Proposed Solution

Enable configuration or something like a `jupyterlab_collaborative_chat_config.py` that is processed on startup, where the path can be specified.

brichet:
Thanks for opening this @mvallet91.

When creating a chat file, you can specify an existing directory in the dialog, like `chats/general.chat` if the `chats` directory already exists.
We could add a setting for a default chat directory, and the path would be relative to this directory.

The left panel also provides a list of existing chat files, as a shortcut to open it (in the left panel).
We have deliberately limited the exploration to the root directory, to avoid extra computation if jupyterlab run in a huge file system.
If we provide a setting for a root chats directory, we should probably update also the directory to explore in the left panel.

mvallet91:
Thank you @brichet. 

So, for `user-a` and `user-b` to have a collaborative chat, they would have to have the same root directory? 

Just to get it right, as I understand it, JupyterHub would start Jupyter in `/home/user-a` and in `/home/user-b`, respectively. And currently, they cannot chat since `jupyter-chat` couldn't read a `/home/shared-chats/CHAT1.chat`, even if they try to create it with that path. Is that correct? Is there another way for them to have access to the same chat?

brichet:
There is special case for collaboration in jupyterhub, you need to set up a collaborative server, but I never tried so I can tell more than that.
https://github.com/jupyterhub/jupyterhub/blob/main/docs/source/tutorial/collaboration-users.md (thanks @trungleduc for the link)

mvallet91:
Thank you, I'll check it out.

PR #91, opened 2024-10-23
Chats directory config

brichet:
Fixes https://github.com/jupyterlab/jupyter-chat/issues/61

Allow the user to configure the default directory for chat creation and list (in side panel).

### Note
This is a suggestion if we wish to continue listing chat files in a single directory in the side panel.
We could also think of a way to manage chat files in the whole tree file:
- look for chat files in large tree file without overload
- display chats coming from several locations (and which can have the same name)
- create chats in different locations


python/jupyterlab-collaborative-chat/schema/factory.json
```diff
@@ -37,6 +37,12 @@
       "default": true,
       "readOnly": false
     },
+    "defaultDirectory": {
+      "description": "Default directory where to create and look for chat.",
+      "type": "string",
+      "default": "",
```
jtpio, line 43: Should the setting provide more information (for example via the `description`) on where files are saved by default? Since an empty string can be ambiguous?

jtpio (approved):
Thanks!

PR #92, opened 2024-10-25
Display chat users

brichet:
This PR adds a toolbar widget to display the connected user on the current chat.

Peek 2024-10-25 16-01

As can be seen in the gif above, displaying all users can take up to 15 seconds (the delay between two user status updates).

It needs to wait for:
- https://github.com/jupyterlab/jupyter-collaboration/pull/379 to be merged and released
- upgrade dependency to `"@jupyter/collaboration": "^3"`


packages/jupyterlab-chat/src/ychat.ts
```diff
@@ -207,6 +207,16 @@ export class YChat extends YDocument<IChatChanges> {
...
+  getMetadata(key: string): IMetadata | undefined {
+    return this._metadata.get(key);
+  }
+
+  setMetadata(key: string, value: PartialJSONValue) {
+    this.transact(() => {
+      this._metadata.set(key, value);
+    });
+  }
+
```
brichet, line 219: The reason why this change has been added in this PR is for a possible usage in `jupyter-ai`:
- the personnas from `jupyter-ai` could be displayed in this toolbar as regular collaborator
- the personna item could be customized to allow disabling it
- disabling a personna could be saved in the metadata for persistence

Even if we do not add the feature described above, adding a way to get/set metadata from a frontend extension may still be useful. But it can be moved to another PR if it make more sense.

PR #93, opened 2024-10-29
Add notes about current development status and breaking changes

brichet:
Initially suggested at https://github.com/jupyterlab/jupyter-chat/pull/91#issuecomment-2441630909


jtpio (approved):
Thanks!

PR #102, opened 2024-11-21
Rename the extension to jupyterlab-chat

brichet:
This PR rename the extension `jupyterlab-chat`, it was `jupyterlab-collaborative-chat` before.

- [x] rename the python and javascript packages
- [ ] rename the directory
- [ ] rename the classes


brichet:
It expected that the check links test fail, the link to the PyPI package does not exist yet.

PR #106, opened 2024-11-25
Typing

brichet:
Adds typing check using mypy on extension.

Also moves the *ui-tests* directory from the python package to the root directory.


brichet:
Failure is not related

Issue #115, opened 2024-12-09
Use dataclass models instead of untyped dictionaries in `YChat` Python API


dlqqq:
### Problem

The `YChat` API expects a dictionary in many areas of the code. For example, the `YChat.get_users()` and `YChat.get_message()` methods return dictionaries, but there's no indication to developers what the types of these dictionaries are, or what keys are available.

### Proposed Solution

Use `dataclass` models instead of dictionaries for the `YChat` Python API. `dataclass` should be preferred over native dictionaries as much as possible on any developer API surface, but the `YChat` API should be prioritized.

Furthermore, the models & attributes within the models should be documented with docstrings. This will make it easier to develop extensions on top of `jupyterlab-chat` if IDE intellisense is available.

krassowski:
Would `TypedDict` or `dataclass` instead work? Adding pydantic closer to core means more dependency conflicts and packaging issues.

dlqqq:
@krassowski Thanks for calling this out. I agree; if we don't need the extra features provided by Pydantic, it seems preferable to use `dataclass`, as it offers a similar API and is native to Python.

I would also add that if we run into issues with `dataclass`, we should feel free to use Pydantic instead. Pydantic v2 is a very common dependency, and it should be safer than many of the dependencies already being pulled in by `jupyterlab-chat`.

ellisonbg:
I think `dataclass` is a reasonable approach for this case.

dlqqq:
I've updated the title & description of this issue accordingly. Thanks folks!

PR #119, opened 2024-12-10
Use dataclass models in YChat

brichet:
Fixes https://github.com/jupyterlab/jupyter-chat/issues/115

The `YChat` uses `dataclass` models for the messages users. The `User` model is inherited (without change) from the jupyter server [model](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/auth/identity.py#L35).
It does not use `dataclass` for metadata, because it can contain anything. Only the `id` field is used by `jupyterlab_chat`.

This PR also save the chat file as a more readable JSON file.

### DIFF
--- a/python/jupyterlab-chat/jupyterlab_chat/models.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/models.py
@@ -1,3 +1,6 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
 from dataclasses import dataclass
 from typing import Literal, Optional
 from jupyter_server.auth import User as JupyterUser
--- a/python/jupyterlab-chat/jupyterlab_chat/models.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/models.py
@@ -1,0 +1,35 @@
+from dataclasses import dataclass
+from typing import Literal, Optional
+from jupyter_server.auth import User as JupyterUser
+
+
+@dataclass(kw_only=True)
+class Message:
+    """ Object representing a message """
+
+    type: Literal["msg"] = "msg"
+    body: str
+    """ The content of the message """
+
+    id: str
+    """ Unique ID """
+
+    time: float
+    """ Timestamp in second since epoch """
+
+    sender: str
+    """ The message sender unique id """
+
+    raw_time: Optional[bool] = None
+    """ Whether the timestamp is raw (from client) or not (from server, unified) """
+
+    deleted: Optional[bool] = None
+    """ Whether the message has been deleted or not (body should be empty if True) """
+
+    edited: Optional[bool] = None
+    """ Whether the message has been edited or not """
+
+
+@dataclass(kw_only=True)
+class User(JupyterUser):
+    """ Object representing a user (same as Jupyter User ) """
--- a/python/jupyterlab-chat/jupyterlab_chat/ychat.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/ychat.py
@@ -185,11 +185,14 @@ class YChat(YBaseDoc):
         Returns the contents of the document.
         :return: Document's contents in JSON.
         """
-        return json.dumps({
-            "messages": self.get_messages(),
-            "users": self.get_users(),
-            "metadata": self.get_metadata()
-        })
+        return json.dumps(
+            {
+                "messages": self.get_messages(),
+                "users": self.get_users(),
+                "metadata": self.get_metadata()
+            },
+            indent=2
+        )
 
     def set(self, value: str) -> None:
         """
--- a/python/jupyterlab-chat/jupyterlab_chat/ychat.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/ychat.py
@@ -12,6 +12,8 @@ from typing import Any, Callable, Optional, Set
 from uuid import uuid4
 from pycrdt import Array, ArrayEvent, Map, MapEvent
 
+from .models import Message, User
+
 
 class YChat(YBaseDoc):
     def __init__(self, *args, **kwargs):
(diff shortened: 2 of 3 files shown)


brichet:
Bumping CI

python/jupyterlab-chat/jupyterlab_chat/models.py
```diff
@@ -0,0 +1,38 @@
...
+    deleted: Optional[bool] = None
+    """ Whether the message has been deleted or not (body should be empty if True) """
+
+    edited: Optional[bool] = None
+    """ Whether the message has been edited or not """
+
+
+@dataclass(kw_only=True)
+class User(JupyterUser):
+    """ Object representing a user (same as Jupyter User ) """
```
dlqqq, line 38: Subclassing this allows for attributes to be added in the future, extending the `User` model provided by Jupyter Server. Good thinking!

python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -52,105 +54,105 @@ def yusers(self) -> Map:
     def ymetadata(self) -> Map:
```
dlqqq, line 54: Does `pycrdt` have generic types, e.g. allowing us to type `ymetadata` as `Map[str, Any]`? 

This would be more helpful for the `ymessages` and `yusers` properties, which are currently typed as just `Array` and `Map` respectively.

python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -52,105 +54,105 @@ def yusers(self) -> Map:
...
+        if not initial_message.id == message.id:
+            initial_message, index = self.get_message(message.id)
             if index is None:
                 return self.add_message(message)
 
         self.update_message(message, index, append)
         return index
 
-    def get_single_metadata(self, name) -> dict:
+    def get_single_metadata(self, name: str) -> dict:
```
dlqqq, line 143: I searched the code and found that this method isn't called anywhere else. Since developers can already safely access a record in the metadata via `ychat.get_metadata().get(key, None)`, can this method be removed?

python/jupyterlab-chat/jupyterlab_chat/models.py
```diff
@@ -0,0 +1,38 @@
...
+    """ The message sender unique id """
+
+    raw_time: Optional[bool] = None
+    """ Whether the timestamp is raw (from client) or not (from server, unified) """
+
+    deleted: Optional[bool] = None
+    """ Whether the message has been deleted or not (body should be empty if True) """
+
+    edited: Optional[bool] = None
+    """ Whether the message has been edited or not """
```
dlqqq, line 33: Do the boolean fields here need to be `Optional`? It seems to make more sense to ensure `raw_time` is set at init, and default to `False` for `deleted` and `edited`. This would allow the PR to be simplified by removing a few `getattr()` calls.

python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -183,11 +185,14 @@ def get(self) -> str:
...
-            "metadata": self.get_metadata()
-        })
+        return json.dumps(
+            {
+                "messages": self.get_messages(),
+                "users": self.get_users(),
+                "metadata": self.get_metadata()
+            },
+            indent=2
+        )
```
dlqqq, line 195: Just curious: does this change make the `*.chat` files more human-readable from a command-line editor? Right now `*.chat` files are a compressed one-line JSON string.

dlqqq (changes requested):

brichet:
Thanks @dlqqq for the review.
This PR needs some updates, the `pycrdt.Array` and `pycrdt.Map` do not allow dataclass.
Sorry, I should have set it as draft. It should not be related to your review, though.

python/jupyterlab-chat/jupyterlab_chat/ychat.py
brichet: Not that I know. But these methods should only be uses to add observers, not to use the data.

Maybe @davidbrochart has a better idea on it.

python/jupyterlab-chat/jupyterlab_chat/ychat.py
brichet: Yes, it is the intention to make it more readable for any text editor.

python/jupyterlab-chat/jupyterlab_chat/models.py
brichet: The idea of using `Optional` was to avoid saving extra information to the file.
I thought it was automatic but not, I added this in https://github.com/jupyterlab/jupyter-chat/pull/119/commits/07b692681454271d09d5e55f811bf03e6f396b86

brichet:
@dlqqq In this PR I changed `YChat.get_messages()`  to "private", because it is not returning the list of `Message` but a list of `dict`, to avoid confusion.

It will collide with https://github.com/jupyterlab/jupyter-ai/pull/115, with make use of this method. I can revert this change.

python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -52,105 +55,119 @@ def yusers(self) -> Map:
...
+        else:
+            return (Message(**message[0]), message[1])
 
-    def get_messages(self) -> list[dict]:
+    def _get_messages(self) -> list[dict]:
         """
         Returns the messages of the document.
         :return: Document's messages.
         """
         return self._ymessages.to_py() or []
```
dlqqq, line 115: > In this PR I changed YChat.get_messages() to "private", because it is not returning the list of Message but a list of dict, to avoid confusion.

Instead of making this method private, could we return a list of `Message` objects? This method is necessary to https://github.com/jupyterlab/jupyter-ai/pull/1151.

```suggestion
    def get_messages(self) -> list[Message]:
        """
        Returns the messages of the document.
        :return: Document's messages.
        """
        message_dicts = self._ymessages.to_py() or []
        return [Message(**message_dict) for message_dict in message_dicts]
```

The performance overhead of this change should be minor compared to the call to `self._ymessages.to_py()`.

dlqqq (changes requested):

python/jupyterlab-chat/jupyterlab_chat/ychat.py
dlqqq: Other calls to `get_messages()` can also be updated accordingly, and will look much cleaner.

python/jupyterlab-chat/jupyterlab_chat/models.py
dlqqq: Sounds reasonable. Can we document the default value for each of these fields in the docstrings?

davidbrochart: BTW you can use `bool | None` instead of `Optional[bool]`, provided that you add this line at the top of the file:
```py
from __future__ import annotations
```

python/jupyterlab-chat/jupyterlab_chat/ychat.py
brichet: Not sure about that, this method will be often used internally without the need of casting every message.
Maybe we could keep this one for internal use, and add a new public method that returns slice of messages as proper `Message` objects.
```python
get_messages(start=0, end=-1, step=1):
```

brichet: I added a public method to get all the messages as a list of Messages.

PR #125, opened 2024-12-16
Remove need for message index in YChat API methods

### LINKED ISSUE #114: Message timestamp flickers `undefined` when streaming
## Description

The timestamp next to each message (e.g. `01:23`) flickers to `undefined` when the message is being frequently updated, e.g. when Jupyter AI streams a reply back to the user.

This is visible from the demo in https://github.com/jupyterlab/jupyter-ai/pull/1043, and I have reproduced this locally on the `v3-dev` branch as well.

### LINKED ISSUE #120: `YChat` should not require an index to update a message
## Description

To update a message in `YChat` (in response to a human user editing a message or Jupyter AI streaming a response), an index is required. See the type signatures for these two methods:

```python
def update_message(self, message: dict, index: int, append: bool = False):
    ...

def set_message(self, message: dict, index: Optional[int] = None, append: bool = False) -> int:
    ...
```

## Proposed solution

1. Add an observer on `self._ymessages` to maintain a lookup table between message IDs and the index in the list. This can be stored in a dictionary whose keys are message IDs and whose values are indices.
2. Update `update_message()` to just accept a `message: Message` argument and an `append: bool` argument. The method should reference the lookup table to find the right index from the message ID.
3. Remove `set_message()`, since developers will always know whether they are adding / updating a message. To make this more obvious, you could define a new dataclass that looks like:

    ```python
    class NewMessage:
        ... # same as `Message`, but without an ID
    ```
    
    Then, update `add_message` to accept `NewMessage` instead of `Message` and return the new message ID. That way, consumers don't have to generate their own UUID, and always know if they're adding/updating a message based on whether they have a message ID or not.

brichet:
## Description

- Fixes #120.
- **EDIT**: also fixes #114.
- Adds a lookup table in YChat, to keep find message index from its ID.
- Adds a `NewMessage` type, containing only the message body and the sender (timestamp and id are generated before inserting the message.
- Removes the `set_message()`, since `add_message()` and `update_message()` do not have the same parameter type anymore.

### DIFF
--- a/python/jupyterlab-chat/jupyterlab_chat/ychat.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/ychat.py
@@ -24,7 +24,7 @@ class YChat(YBaseDoc):
         self._ydoc["users"] = self._yusers = Map()
         self._ydoc["messages"] = self._ymessages = Array()
         self._ydoc["metadata"] = self._ymetadata = Map()
-        self._ymessages.observe(self._timestamp_new_messages)
+        self._ymessages.observe(self._on_messages_change)
 
         # Observe the state to initialize the file as soon as the document is not dirty.
         self._ystate_subscription = self._ystate.observe(self._initialize)
@@ -132,17 +132,10 @@ class YChat(YBaseDoc):
 
         with self._ydoc.transaction():
             index = len(self._ymessages) - next((i for i, v in enumerate(self._get_messages()[::-1]) if v["time"] < timestamp), len(self._ymessages))
-            appended = index == len(self._ymessages)
             self._ymessages.insert(
                 index,
                 asdict(message, dict_factory=message_asdict_factory)
             )
-
-        if appended:
-            self._indexes_by_id[uid] = index
-        else:
-            # if the message has been inserted in the list,
-            self._indexes_by_id = {message["id"]: idx for idx, message in enumerate(self._get_messages())}
 
         return uid
 
@@ -231,9 +224,7 @@ class YChat(YBaseDoc):
                     self._yusers.update({k: v})
 
             if "messages" in contents.keys():
-                for message in contents["messages"]:
-                    self._indexes_by_id[message["id"]] = len(self._ymessages)
-                    self._ymessages.append(message)
+                self._ymessages.extend(contents["messages"])
 
             if "metadata" in contents.keys():
                 for k, v in contents["metadata"].items():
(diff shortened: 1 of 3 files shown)


python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -109,63 +112,52 @@ def get_messages(self) -> list[Message]:
...
                 asdict(message, dict_factory=message_asdict_factory)
             )
-            return index
 
-    def update_message(self, message: Message, index: int, append: bool = False):
+        if appended:
+            self._indexes_by_id[uid] = index
+        else:
+            # if the message has been inserted in the list,
+            self._indexes_by_id = {message["id"]: idx for idx, message in enumerate(self._get_messages())}
```
dlqqq, line 145: Consider this potential scenario:

- Two frontend Yjs clients (A & B) open an empty chat.
- Both clients write a new message at the same time, but client B has an extra 500ms of latency.
- A's update appends a new item to an empty list, so it is assigned index 0.
- Later, B's update also appends a new item to an empty list, so it is also assigned index 0. Therefore A's index gets bumped from 0 to 1.

I believe this scenario would lead to a bug, where `self._indexes_by_id` would track the index of A's message as `0` instead of `1`.

However, I'm not sure if this scenario would actually happen. I'm OK with this PR being merged with the existing logic, but I think we should also explore if it's more reliable to maintain `self._indexes_by_id` via an observer that listens to every update to the YDoc. That way, it can update the lookup table in response to updates triggered by frontend clients, not just other server extensions calling the Python API.

Let's chat about this more at standup. 👍

dlqqq (changes requested):
@brichet This looks great. Thank you for adding unit tests & for adding the `NewMessage` dataclass! ❤️ 

I've left a few discussion topics below to review at our technical standup tomorrow. 👋

brichet:
I rework it to index also messages from frontend.

Sorry @dlqqq for the first review.

python/jupyterlab-chat/jupyterlab_chat/ychat.py
brichet: > it's more reliable to maintain self._indexes_by_id via an observer that listens to every update to the YDoc. That way, it can update the lookup table in response to updates triggered by frontend clients, not just other server extensions calling the Python API.

IIUC this is what does the last commit, by updating the index on any change in the `ymessages`.

dlqqq (approved):
@brichet Awesome, TYSM! 🎉

PR #126, opened 2024-12-16
Prevent sending empty message

### LINKED ISSUE #118: Keyboard sends empty messages when input is empty
## Description

Pressing Enter/Shift-Enter sends empty messages when input is empty, despite the "Send" button being disabled.

## Reproduce

On `v3-dev` branch of Jupyter AI, press Enter/Shift-Enter repeatedly while the input is empty.

https://github.com/user-attachments/assets/b93d5e40-72bb-4264-84b6-a6a55fbc16ff

## Expected behavior

- Enter/Shift-Enter should not send a message when the input is empty.
- When Enter/Shift-Enter is pressed:
  - Determine the message is all whitespace via regex, and if so, clear the input & do nothing else.
  - Otherwise, send the message.
  - The send button should be enabled/disabled according to this state. There should not be 2 separate state vars that control whether the send button is enabled & control whether the input should be sent.

brichet:
Fixes #118 

Port https://github.com/jupyterlab/jupyter-ai/pull/946

**EDIT**
- prevent sending empty message (only whitespace) with keyboard and button
- remove condition on 'cancel' button to disable it

### DIFF
--- a/packages/jupyter-chat/src/components/chat-input.tsx
+++ b/packages/jupyter-chat/src/components/chat-input.tsx
@@ -130,16 +130,22 @@ export function ChatInput(props: ChatInput.IProps): JSX.Element {
       return;
     }
 
-    // do not send the message if the user was selecting a suggested command from the
+    // Do not send the message if the user was selecting a suggested command from the
     // Autocomplete component.
     if (highlighted) {
       return;
     }
 
+    // Do not send empty messages, and avoid adding new line in empty message.
+    if (!input.trim()) {
+      event.stopPropagation();
+      event.preventDefault();
+      return;
+    }
+
     if (
-      event.key === 'Enter' &&
-      ((sendWithShiftEnter && event.shiftKey) ||
-        (!sendWithShiftEnter && !event.shiftKey))
+      (sendWithShiftEnter && event.shiftKey) ||
+      (!sendWithShiftEnter && !event.shiftKey)
     ) {
       onSend();
       event.stopPropagation();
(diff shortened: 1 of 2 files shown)


packages/jupyter-chat/src/components/chat-input.tsx
```diff
@@ -130,16 +130,22 @@ export function ChatInput(props: ChatInput.IProps): JSX.Element {
...
     if (highlighted) {
       return;
     }
 
+    // Do not send empty messages, and avoid adding new line in empty message.
+    if (!input.trim()) {
+      event.stopPropagation();
+      event.preventDefault();
+      return;
+    }
```
dlqqq, line 144: This handler only sends the message if `!input.trim()`, but the `SendButton` component is enabled by a different condition (when `input.length >= 0`).

```tsx
<SendButton
  model={model}
  sendWithShiftEnter={sendWithShiftEnter}
  inputExists={input.length > 0}
  onSend={onSend}
  hideIncludeSelection={hideIncludeSelection}
  hasButtonOnLeft={!!props.onCancel}
/>

```

Currently, if the input contains only white space, the send button will be enabled (which shouldn't happen), but pressing Enter/Shift-Enter does nothing. This is a difference in behavior caused by using two separate state variables.

I recommend defining a new local variable `const inputExists = !!input.trim()` in the component's top scope, and using that variable in both the handler & the send button.

dlqqq (changes requested):
@brichet Thanks! One minor comment below.

packages/jupyter-chat/src/components/chat-input.tsx
brichet: Good catch, thanks

brichet: Updated in https://github.com/jupyterlab/jupyter-chat/pull/126/commits/b28a6983207a22119ef03ca4f5ea251ef62c1252

brichet:
The condition to disable the button was also applied to the 'cancel' button (visible when editing a message).
I removed this condition, since an edition should always be cancellable, AFAIK.

dlqqq (approved):
@brichet TY! Great work

PR #129, opened 2024-12-19
Fix math rendering using Latex delimiters

### LINKED ISSUE #110: LaTeX delimiters can't render math in chat
## Description

The current UI doesn't render math markup enclosed in LaTeX delimiters. There are two types of LaTeX delimiters:

- Inline math: `\( <math> \)`
- Display math: `\[ <math> \]`

## Reproduce

1. Use latest `jupyterlab-chat`.
2. Send a message: `Given \(x = 2\), what is \(x + 2\)?`
3. Verify that math markup doesn't work in Jupyter Chat, but works in Jupyter AI.

The below screenshot has Jupyter AI on the left, and Jupyter Chat on the right.

Screenshot 2024-12-04 at 10 05 51 AM

## Additional context

The fix is unclear; we need to do further research. The PRs that affect math rendering were ported properly as far as I can tell:

- https://github.com/jupyterlab/jupyter-chat/pull/95
- https://github.com/jupyterlab/jupyter-chat/pull/99

I think the frontend implementation needs to be modified to make sure this function is called on update: https://github.com/brichet/jupyter-chat/blob/7bcdd868cddbf26268e2a4147b4ad4f32d4128ca/packages/jupyter-chat/src/components/rendermime-markdown.tsx#L26-L40

brichet:
Fixes #110 

image

### DIFF
--- a/packages/jupyter-chat/src/components/rendermime-markdown.tsx
+++ b/packages/jupyter-chat/src/components/rendermime-markdown.tsx
@@ -33,10 +33,10 @@ type RendermimeMarkdownProps = {
  */
 function escapeLatexDelimiters(text: string) {
   return text
-    .replace('\\(/g', '\\\\(')
-    .replace('\\)/g', '\\\\)')
-    .replace('\\[/g', '\\\\[')
-    .replace('\\]/g', '\\\\]');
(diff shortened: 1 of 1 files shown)


dlqqq (approved):
@brichet Thank you!

PR #132, opened 2024-12-20
Export components from @jupyter/chat

brichet:
Exports the components from the package `@jupyter/chat`.

This will allow reusing it in other applications.

It also renames the `RendermimeMarkdown` to `MarkdownRenderer` for clarification.


brichet:
Bump CI

brichet:
Merging this one since it mostly about adding components to export.

PR #146, opened 2025-01-07
Adds side panel widgets to the tracker

### LINKED ISSUE #130: Add tracker for chats in side panel
### Problem

The chats opened in the side panel are not tracked, unlike the ones in main area.

This leads to 2 issues (at least):
- the side panel is not restored when reloading the page
- the shortcut to focus the chat input  cannot handle the side panel chat

### Proposed Solution

Adds the chat in side panel to the widget tracker.

### LINKED ISSUE #139: Keyboard shortcuts: fix warnings and add one for focusing chat input
A few minor issues with keyboard shortcuts with Jupyter AI 3.0.0a0 pre-release.

## Description

First, I see the following warnings in the browser console when using Jupyter AI 3.0.0a0 pre-release:

Screenshot 2024-12-30 at 10 18 34 AM

Second, when I have a chat document open and active, pressing enter doesn't focus the text box like I would expect.

## Reproduce

1. Open chat document.
2. Make sure the text box is blurred, while the document is active.
3. Press enter to focus the text box.
4. See that it doesn't get focus (mouse click is required).

## Expected behavior

* As a user, I want to use enter to focus the text box of a chat document.

brichet:
Add the side panel chats to the widget tracker.

Fixes #130
Fixes #139

The chats tracker tracks both `LabChatPanel` (the main area widget) and `ChatWidget` (widget in the side panel).

#### IMPROVEMENT

- The layout restorer is now able to open back the chats in the main area and in the side panel.
- The shortcut to focus on the chat input (`Accel Shift 1`) works for all the chat widgets. The last used input is focused.
- EDIT: the chat input is focussed when clicking anywhere on the chat except on a focusable element.

#### TO DO
- [x] improve the tracking of the chat widget
  The widget become the current widget of the tracker only if an input has been clicked. We should focus on the input component when the chat is clicked (anywhere except on an other input). This would also be a nice UX improvement.
- [x] add tests on it


packages/jupyterlab-chat/src/token.ts
```diff
@@ -56,7 +57,7 @@ export interface IChatFactory {
   /**
    * The chat panel tracker.
    */
-  tracker: IWidgetTracker<LabChatPanel>;
+  tracker: WidgetTracker<LabChatPanel | ChatWidget>;
```
jtpio, line 60: Do we need to expose a full `WidgetTracker`? Or could we stick to the more limited `IWidgetTracker` as before?

packages/jupyterlab-chat/src/widget.tsx
```diff
@@ -152,16 +152,13 @@ export class ChatPanel extends SidePanel {
    * @param model - the model of the chat widget
    * @param name - the name of the chat.
    */
-  addChat(model: IChatModel, path: string): void {
+  addChat(model: IChatModel): ChatWidget {
```
jtpio, line 155: Note: this will likely require a `0.8.0` release.

packages/jupyterlab-chat/src/token.ts
dlqqq: +1

packages/jupyter-chat/src/widgets/chat-widget.tsx
```diff
@@ -14,11 +14,12 @@ export class ChatWidget extends ReactWidget {
   constructor(options: Chat.IOptions) {
     super();
 
-    this.id = 'jupyter-chat::widget';
     this.title.icon = chatIcon;
     this.title.caption = 'Jupyter Chat'; // TODO: i18n
 
     this._chatOptions = options;
+    this.id = `jupyter-chat::widget::${options.model.name}`;
+    this.node.onclick = () => this.model.focusInput();
```
dlqqq, line 22: Note that this change causes the input to essentially always be focused unless the user clicks out of the chat widget. I think this is fine, but want us to be aware of this change.

python/jupyterlab-chat/src/index.ts
```diff
@@ -298,11 +299,24 @@ const docFactories: JupyterFrontEndPlugin<IChatFactory> = {
...
-        command: 'docmanager:open',
-        args: panel => ({ path: panel.context.path, factory: FACTORY }),
-        name: panel => panel.context.path,
-        when: app.serviceManager.ready
+        command: CommandIDs.openChat,
+        args: widget => ({
+          filepath: widget.model.name ?? '',
+          inSidePanel: widget instanceof ChatWidget
+        }),
+        name: widget => widget.model.name,
```
dlqqq, line 318: I spent about 15 min reviewing how this PR works, and I'd like to confirm my understanding. This works because:

- The tracker maintains a list of all chat widgets, now including widgets in the side panel.
- This list is ordered by usage, i.e. the last item in this list is the most recently opened chat widget.
- When this plugin is activated, the layout restorer iterates through this list and opens each chat widget, automatically focusing the most recently opened chat widget.

Is this correct?

dlqqq (approved):
@brichet Looks good, thank you for building this! Tested your PR locally and it seems to work well. 🎉 Just left a few extra comments which don't block this PR.

packages/jupyter-chat/src/widgets/chat-widget.tsx
brichet: Yes, actually the `WidgetTracker` relies on the [`FocusTracker`](https://github.com/jupyterlab/lumino/blob/main/packages/widgets/src/focustracker.ts), which needs to have an element focused to trigger the change of current widget.

This is mentioned in the PR description:
> EDIT: the chat input is focussed when clicking anywhere on the chat except on a focusable element.

python/jupyterlab-chat/src/index.ts
brichet: Yes, this is how it work. The chats are restored in order, whether they are in the side panel or the main area.

packages/jupyterlab-chat/src/token.ts
brichet: Yes, this is required in the current implementation, because the side panel uses the `add` method (not exposed in the interface) when creating a widget.
```ts
factory.tracker.add(widget);
```

Issue #147, opened 2025-01-08
Design a way to attach JSON objects to messages


dlqqq:
### Problem

There exists a need to design some way of attaching arbitrary JSON objects to messages.

For Jupyter AI v3, we intend to provide `@` commands that allow users to include files & variables from the notebook for context. It would be ideal to retrieve the data of these files/variables directly from the frontend and somehow attach them to the next input for the current user. Defining command handling in the frontend is the most flexible approach, as it allows access to every API provided by Jupyter, not just Python APIs. See this [comment](https://github.com/jupyterlab/jupyter-chat/issues/143#issuecomment-2569923586) for details on the thought process here.

Attachments can also later be used to add support for multimodal models which accept both text & images, and may return text & images.

### Proposed Solution

1. Modify the structure of `Message` to allow a list of attachments, whose contents are arbitrary and JSON-serializable. 
2. Design some sort of **Attachments API** in the frontend.
3. Add a UI to let users attach files manually. This should call the Attachments API.
4. Show attachments as a list of toasts on top of the chat input, and allow users to remove attachments. I've included an example screenshot from GitHub Copilot to use as inspiration.

<img width="126" alt="Screenshot 2025-01-07 at 5 22 06 PM" src="https://github.com/user-attachments/assets/b2a528e6-7f17-4012-92ff-37f7bb8693e9" />

I'm not sure how the Attachments API should be accessed. I think it would be good to make it accessible from the `YChat` object in the frontend. This way, the `YChat` interface provides all of the APIs needed by chat commands.

### Additional context

I'm currently quite busy with iterating on my existing designs & planning collaborations with new contributors. @brichet I'm assigning this to you as you have the most context on Jupyter Chat, and can help propose better ideas. Feel free to implement a proof-of-concept in a separate PR.

brichet:
Thanks @dlqqq for opening this, I can take a look at it.

#### Proposition

- add an `attachments` property to `Message`, containing an array of objects, with this kind of interface (can be extended):

```ts
{
  type: 'file' | 'variable' | 'image',
  value: string,
  mimetype?: string
}
```
The value could be a path, a variable name or an image content.

- add a function in [`ChatInput`](https://github.com/jupyterlab/jupyter-chat/blob/8de473aa176a5fbdf8ff7165ff629606b5d36a74/packages/jupyter-chat/src/components/chat-input.tsx#L31) to add an attachment to a list, that will be send with the message (can be called several time to add several attachments to one message).

[...]

PR #148, opened 2025-01-17
Implement message attachments

brichet:
## Description

- Implements new methods for adding, removing, and viewing message attachments.
- Fixes https://github.com/jupyterlab/jupyter-chat/issues/147.

## Demo

[record-2025-02-13_11.04.21.webm](https://github.com/user-attachments/assets/aabe51c1-2054-4837-af39-d74548d9962f)

There is now a lite deployment in the built doc including a simple extension with a chat panel. I updated it to allow attachments https://jupyter-chat--148.org.readthedocs.build/en/148/lite/lab/index.html


packages/jupyter-chat/src/model.ts
```diff
@@ -166,6 +172,26 @@ export interface IChatModel extends IDisposable {
...
+
+  /**
+   * Remove attachment to the next message to send.
+   */
+  removeAttachment?(attachment: IAttachment): void;
+
+  /**
+   * Update attachments.
+   */
+  updateAttachments?(attachments: IAttachment[]): void;
```
dlqqq, line 188: This only gets used to clear the list of attachments, and there isn't a clear use-case for bulk updates. Given this, can you rename this method to `clearAttachments()` and update the implementation accordingly?

If adding/removing multiple files is required in the future, that capability should be provided by new `addAttachments()` and `removeAttachments()` methods which accept a list of attachments to add/remove.

packages/jupyter-chat/src/model.ts
```diff
@@ -166,6 +172,26 @@ export interface IChatModel extends IDisposable {
...
+
+  /**
+   * Open attachments.
+   */
+  clickAttachment?(attachments: IAttachment): void;
```
dlqqq, line 193: Is this the responsibility of the chat model? Opening an attachment doesn't affect chat state, so IMO this method doesn't need to be defined in `IChatModel`. 

Can you explore modifying `AttachmentsComponent` to handle opening attachments directly through the command registry? Then this method and its implementation in `ChatModel` can be removed.

packages/jupyter-chat/src/model.ts
```diff
@@ -520,6 +553,53 @@ export class ChatModel implements IChatModel {
...
+    );
+    if (index > -1) {
+      this.inputAttachments.splice(index, 1, attachment);
+    } else {
+      this.inputAttachments.push(attachment);
+    }
+    if (emit) {
+      this._inputAttachmentsChanges.emit([...this.inputAttachments]);
+    }
+  };
```
dlqqq, line 571: 1. Is there a need for an `emit` argument if we usually default to `true`? IMO, allowing consumers to not emit to `inputAttachmentsChanges` defeats the purpose of also allowing consumers to listen to that signal.


2. Can this logic be simplified further? The ordering of attachments should not matter, so we can return early if a duplicate attachment already exists. For example:


```suggestion
  addAttachment = (attachment: IAttachment): void => {
    const duplicateAttachment = this.inputAttachments.find(
      att => att.type === attachment.type && att.value === attachment.value
    );
    if (duplicateAttachment) {
      return;
    }
    
    this.inputAttachments.push(attachment);
    this._inputAttachmentsChanges.emit([...this.inputAttachments]);
  };
```

packages/jupyter-chat/src/model.ts
```diff
@@ -520,6 +553,53 @@ export class ChatModel implements IChatModel {
...
+   */
+  removeAttachment = (attachment: IAttachment): void => {
+    const index = this.inputAttachments.findIndex(
+      att => att.type === attachment.type && att.value === attachment.value
+    );
+    if (index > -1) {
+      this.inputAttachments.splice(index, 1);
+    }
+    this._inputAttachmentsChanges.emit([...this.inputAttachments]);
+  };
```
dlqqq, line 584: I think that similarly here, it makes sense to only emit the signal when the list changes:


```suggestion
  removeAttachment = (attachment: IAttachment): void => {
    const attachmentIndex = this.inputAttachments.findIndex(
      att => att.type === attachment.type && att.value === attachment.value
    );
    if (attachmentIndex === -1) {
      return;
    }
    
    this.inputAttachments.splice(attachmentIndex, 1);
    this._inputAttachmentsChanges.emit([...this.inputAttachments]);
  };
```

dlqqq (changes requested):
@brichet Thank you for working on this while I was away! Left some feedback below.

packages/jupyter-chat/src/components/attachments.tsx
```diff
@@ -0,0 +1,88 @@
...
+const ATTACHMENT_CLICKABLE_CLASS = 'jp-chat-attachment-clickable';
+const REMOVE_BUTTON_CLASS = 'jp-chat-attachment-remove';
+
+/**
+ * The attachments props.
+ */
+export type AttachmentsProps = {
+  attachments: IAttachment[];
+  onClick?: (attachment: IAttachment) => void;
+  onRemove?: (attachment: IAttachment) => void;
```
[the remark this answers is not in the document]
brichet: Actually no, the `onRemove` is available only on the input, not on the message itself. Currently it is not possible to remove an attachment from an already sent message. We should add this when updating a message.

I also think that `onClick` property should be optional for future kind of attachments that may not have a specific preview on click (e.g. variables).

packages/jupyter-chat/src/types.ts
```diff
@@ -44,12 +44,13 @@ export interface IConfig {
...
  * The chat message description.
  */
-export interface IChatMessage<T = IUser> {
+export interface IChatMessage<T = IUser, U = IAttachment> {
   type: 'msg';
   body: string;
   id: string;
   time: number;
   sender: T;
+  attachments?: U[];
```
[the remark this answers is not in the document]
brichet: This interface is reused in `jupyterlab-chat` packages with the `IYmessage`. In this case, it is only an id of an attachment (same as the users), to avoid saving several time the same attachment in the final document.

packages/jupyterlab-chat/src/model.ts
```diff
@@ -150,6 +157,16 @@ export class LabChatModel extends ChatModel implements DocumentRegistry.IModel {
...
+
+    // Add the attachments to the message.
+    if (this.inputAttachments.length) {
+      const attachmentIds = this.inputAttachments.map(attachment =>
+        this.sharedModel.setAttachment(attachment)
+      );
+      msg.attachments = attachmentIds;
+      this.updateAttachments([]);
+    }
+
```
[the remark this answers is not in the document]
brichet: Right, thanks for the suggestion.
Indeed, the `ChatModel` class in `@jupyter/chat` is not intended to be instantiated, but only to be extended.
I opened https://github.com/jupyterlab/jupyter-chat/issues/157.

brichet:
Thanks for the review @dlqqq, I addressed some of your comment. I'll take a look at an alternative to `clickAttachments()`.

packages/jupyter-chat/src/model.ts
brichet: Agree, in addition a consumers of the chat widget or the jupyterlab-chat extension should be able to handle other attachment types to other actions / commands.

An option could be to add a registry to the chat widget, linking the function to be called according to the type of attachment (like a `Map<string, (attachment: IAttachment) => void>`).
There could be a token to expose this registry from `jupyterlab-chat`, to allow external extensions (e.g. `jupyter-ai`) to register new behavior with new attachment type (like `var`).

If the registry is not provided with the chat widget options, it could lead to a default one that can open files (as soon as a command registry is provided, at least).

dlqqq: Would it be helpful to build some kind of `JupyterlabContext` React context which provides access to all the APIs provided by other JupyterLab extensions?

For example, I was thinking that a component could call something like:


```ts
const { commandRegistry } = useJupyterlabContext();
```

and similarly so to access other JupyterLab singletons. This would simplify the frontend code by not requiring us to pass every singleton as a prop to every child component. This is also equivalent in terms of performance, as these singleton objects are not changed at runtime.

brichet: @dlqqq I added a registry to add opener function for a given attachment type.
There is a default opener for  `file` type provided by `jupyterlab-chat`, which is available with a token to register more types in external extensions.

As you suggested, the registry is passed back to the react components using a context.
Indeed, we should extend the context to other singletons, probably all the props in [`IChatBodyProps`](https://github.com/jupyterlab/jupyter-chat/blob/73512c16a6c602b81f873ecf9a3c20cb73cefd94/packages/jupyter-chat/src/components/chat.tsx#L111)

packages/jupyter-chat/src/model.ts
```diff
@@ -91,6 +92,11 @@ export interface IChatModel extends IDisposable {
    */
   readonly focusInputSignal?: ISignal<IChatModel, void>;
 
+  /**
+   * A signal emitting when the input attachments changed.
+   */
+  readonly inputAttachmentsChanges?: ISignal<IChatModel, IAttachment[]>;
```
jtpio, line 98: Maybe with the following spelling, similar to the names of other signals?

```suggestion
  readonly inputAttachmentsChanged?: ISignal<IChatModel, IAttachment[]>;
```

brichet:
> The recurrent failing test do not fails locally. And strangely, it constantly fails at opening the chat, which is done in other tests with the same chat file...

The failing test is the first one of the notification tests. After reordering the tests, the 'new' first test is now failing.
I'll try to fix it, otherwise we would have to disable the whole test group.

brichet:
I finally skipped the notification test for now, I'll open a issue when this PR is merged to restore it in a follow up PR.

packages/jupyter-chat/src/types.ts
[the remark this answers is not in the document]
dlqqq: Ah, I see the mention in `packages/jupyterlab-chat/src/ychat.ts`. I'm OK with this for now, but I also think that it would be simpler to just have the two be separate types.

Generic types can cause a lot of complex TypeScript errors that are difficult to resolve, so I recommend we avoid them unless they are necessary. This seems equivalent & easier to maintain:

```suggestion
interface IBaseChatMessage {
  type: 'msg';
  body: string;
  id: string;
  time: number;
  sender: unknown;
  attachments?: unknown[];
}

/**
 * The interface of a chat message in the chat model.
 */
export interface IChatMessage extends IBaseChatMessage {
  sender: IUser;
  attachments?: IAttachment[];
}

/**
 * The interface of a chat message in the YChat shared document.
 */
export interface IYMessage extends IBaseChatMessage {
  sender: string;
  attachments?: string[];
}
```

dlqqq (changes requested):
Thank you @brichet, the new changes look great! Left some additional minor feedback.

dlqqq (approved):
I'll approve this PR because none of my latest feedback is worth blocking this PR for another day.

If you do merge, please open follow-up issues for my latest feedback, thank you!

PR #158, opened 2025-02-07
Allow use of up and down arrow in chat input

### LINKED ISSUE #155: Not able to use the up and down arrows
## Description

Currently it does not seem to be possible to use the <kbd>↑</kbd> and <kbd>↓</kbd> arrows in the chat input area.

## Reproduce

Going up and down is often useful, for example when trying to type a code block with
<pre>
```

```
</pre>

https://github.com/user-attachments/assets/06fa794c-13fc-4a8d-b456-b5687bea01fc

## Expected behavior

It should be possible to navigate through the text input with the up and down arrow keys (left and right seem to be working fine)

## Context

Tested on the latest commit: ff010157cf6dc6b691367cfac4971b685b2f169e

brichet:
Fixes #155 

This PR prevent propagating the up and down arrow event to the `Autocomplete` component if the autocompletion is not opened.
It is weird that it is not handled by material UI itself, maybe there is a better solution.

### DIFF
--- a/packages/jupyter-chat/src/components/chat-input.tsx
+++ b/packages/jupyter-chat/src/components/chat-input.tsx
@@ -128,6 +128,11 @@ export function ChatInput(props: ChatInput.IProps): JSX.Element {
   }, [input]);
 
   function handleKeyDown(event: React.KeyboardEvent<HTMLInputElement>) {
+    if (['ArrowDown', 'ArrowUp'].includes(event.key) && !open) {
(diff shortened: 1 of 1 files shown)


brichet:
cc. @dlqqq to port it to `jupyter-ai` v2, since it may fix https://github.com/jupyterlab/jupyter-ai/issues/939

dlqqq:
Thanks! I've assigned it to a new member on our team.

PR #159, opened 2025-02-10
Update eslint rules to avoid importing MUI icons

brichet:
This PR ports https://github.com/jupyterlab/jupyter-ai/pull/1225 by updating the `eslint` rules, to add enforcement rules for path imports on mui (thanks @krassowski).

It also moves all the linting configurations to dedicated files, instead of having all of them in the `package.json` file, for a better readability.


dlqqq:
@brichet I can review this once the UI tests pass. Thank you for opening this!

brichet:
Thanks @dlqqq, it passes now.
We should fix these flaky tests.

dlqqq (approved):
@brichet Great work, thank you! 🎉 I verified that `jlpm eslint:check` fails when importing MUI icons incorrectly, and that this is being run in CI (specifically in `Build / Build jupyter_chat`).

PR #160, opened 2025-02-10
Fix disabled code toolbar buttons

### LINKED ISSUE #154: Disabled "insert above" and "insert below" buttons (no active cell)
## Description

In some situations the buttons to insert code snippets as new cells in a notebook seem to be disabled while they shouldn't.

## Reproduce

Create a new notebook and click on the `cut` icon in the toolbar several times.

With `jupyterlab-chat`:

https://github.com/user-attachments/assets/dda4507d-72d3-483a-9cae-a9b3b36e1bff

With an extension using `buildChatSidebar` from `@jupyter/chat`:

https://github.com/user-attachments/assets/3b7e3a29-f265-4d74-b245-bb43678e231f

## Expected behavior

The buttons should always be enabled when there is a notebook with the focus.

brichet:
Fixes #154 by removing a concurrency listener on cell disposed, which was colliding with the listener on cell changed.
I can't remember why it was added the first time, but it has been done in a dedicated commit so maybe it was fixing another corner case. I was not able to reproduce it.

Working on this in dark mode make me realized that the svg paths were missing the `jp-chat3` class, that is used to handle the theme changes. This PR also fixes it.

__BEFORE__ (with all buttons enabled):
image

__AFTER__ (with all buttons enabled):
image


jtpio (approved):
Thanks!

Looking good:

https://github.com/user-attachments/assets/4e84fe29-cc79-4b4e-82b1-7296061f2884

Issue #144, opened 2025-01-04
 Design proposal: Chat Completions API (rev. 1.1)


dlqqq:
## Description

This issue proposes a design for a new **Chat Completions API**. This API will allow *consumer extensions* to provide completions for the user's current input from the UI. In this context, a consumer extension is any frontend + server extension that intends to provide completions for substrings in the chat input.

- cc @ellisonbg @mlucool @michaelchia

- Supersedes #143.

### Motivation

Suppose a user types `/` in the chat with Jupyter AI installed. Today, Jupyter Chat responds by showing a menu of **chat completions**:

<img width="358" alt="Screenshot 2025-01-02 at 5 08 27 PM" src="https://github.com/user-attachments/assets/833baa29-3a93-4943-aac7-1df8299496fd" />

The opening of this completions menu is triggered simply by typing `/`. However, the current implementation only allows a single "trigger character" (`/`). This means that `@` commands in Jupyter AI cannot be autocompleted. Furthermore, the completer makes a network call every time a user types `/`.

This design aims to:

1. Allow multiple completers to provide completions,
2. Allow triggering patterns to be strictly defined, and
3. Generate completions in a way which minimizes network calls.

To help explain the proposed design, this document will start from the perspective of a consumer extension, then work backwards towards the necessary changes in Jupyter Chat.

## Step 1: Define a new `IChatCompleter` interface

To register completions for partial inputs, a consumer extension must provide a set of **chat completers**. A chat completer is a JavaScript/TypeScript class which provides:

- `id` (property): Defines a unique ID for this chat completer. We will see why this is useful later.

- `regex` (property): Defines a regex which matches any incomplete input.
  - Each regex should end with `$` to ensure this regex only matches partial inputs just typed by the user. Without `$`, the completer may generate completions for commands which were already typed.
  
- `async initialize(): void`: called and awaited by Jupyter Chat.

- `async getCompletions(match: str): ChatCompletion[]`: Defines a method which accepts a substring matched by its regex, and returns a list of potential completions for that input. This list may be empty.
  - It would be helpful to think of this method as just returning a list of completions for the user's input. We return a list of objects to allow each completion to have metadata, such as the description & icon.

It's important to note that a consumer extension may provide more than 1 completer. This allows extensions to provide completions for different commands which aren't easily captured by a single regex. For example, Jupyter AI can have a completer for `/` commands and another completer for `@` commands.

[...]

PR #161, opened 2025-02-11
[WIP] Implement chat commands

dlqqq:
## Description

Implements a new system of chat commands in Jupyter Chat.

This PR is very much a work-in-progress. Right now, this branch provides a minimal working example of the design proposed in #144. CI is expected to fail for now.

## Demo

The chat commands API is now much more expressive and extensible. This design allows for multiple command providers, which can each provide a unique set of commands.

See this demo, where 2 custom providers are used to provide two sets of commands: one for slash commands, and one for emojis. Note that when an emoji command is invoked, it replaces the partial input with the emoji. This demonstrates how the new design can provide a more intuitive UX by allowing for operations previously unavailable in Jupyter Chat and Jupyter AI.

https://github.com/user-attachments/assets/073206cf-8a54-4889-af75-5c9d3e7657a5


python/jupyterlab-chat/src/chat-commands/slash-command-demo.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    replaceCurrentWord(command.replaceWith);
+  }
+}
+
+export const slashCommandDemoPlugin: JupyterFrontEndPlugin<void> = {
+  id: 'jupyterlab-chat-extension:slashCommandDemoPlugin',
+  description:
+    'A demo plugin which adds Jupyter AI slash commands to the menu. Should be removed after PR review.',
```
jtpio, line 52: Maybe this could be moved to the JupyterLite demo used in the docs?

https://github.com/jupyterlab/jupyter-chat/tree/main/docs/jupyter-chat-example

packages/jupyter-chat/src/components/input/use-chat-commands.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+        const replaceCurrentWordClosure = (newWord: string) => {
+          replaceCurrentWord(input, cursorIndex, newWord, setInput);
+        };
+
+        chatCommandRegistry.handleChatCommand(
+          command,
+          currentWord,
+          replaceCurrentWordClosure
+        );
```
brichet, line 140: Couldn't we replace the word in the `onChange()` callback directly. Since we have the command object, we could check if the attribute `replaceWith` is defined and replace the string. 

```ts
if (command.replaceWith) {
  replaceCurrentWord(input, cursorIndex, command.replaceWith, setInput);
}
```

Is there any advantage to replace it from the `handleChatCommand()` function ?

packages/jupyter-chat/src/chat-commands/types.ts
```diff
@@ -0,0 +1,70 @@
...
+   * commands menu.
+   *
+   * TODO: Pass a ChatModel/InputModel instance here to provide a function to
+   * replace the current word.
+   */
+  handleChatCommand(
+    command: ChatCommand,
+    currentWord: string,
+    replaceCurrentWord: (newWord: string) => void
+  ): Promise<void>;
```
brichet, line 69: From the comment above, this interface would be reduced.
```suggestion
  handleChatCommand(
    command: ChatCommand
  ): Promise<void>;
```

but maybe I missed some use cases.

packages/jupyter-chat/src/components/input/use-chat-commands.tsx
```diff
@@ -0,0 +1,168 @@
...
+         * On close: set `highlighted` to `false` and close the popup by
+         * setting `open` to `false`.
+         */
+        () => {
+          setHighlighted(false);
+          setOpen(false);
+        }
+    },
+    menu: {
+      open,
```
brichet, line 164: The `open` state is returned 2 times, in the `autocompleteProps` and in the `menu` props. Don't know if this expected, but both are used in the input component.

packages/jupyterlab-chat/src/factory.ts
```diff
@@ -7,6 +7,7 @@ import {
   ChatWidget,
   IActiveCellManager,
   IAutocompletionRegistry,
```
brichet, line 9: We should probably remove all the references to `IAutocompletionRegistry`, `IAutocompletionCommandsProps` and maybe others, since this PR is a replacement to it.

python/jupyterlab-chat/src/chat-commands/registry.ts
brichet, line 1: This file could live in `@jupyter/chat` package as a default registry that could be used by any extension, I think.

brichet (changes requested):
Thanks @dlqqq for working on it.

I left some comments below.

brichet:
In a more general test, I saw 2 issues in usages, that can be seen in the following screencast:
- after replacing an input text using a commands, a new line replace again the text with `undefined`
- the completion menu does not open if the word is not at the beginning of the input. Strangely, when printing the `autocompleteProps`, the `open` props is set to `true`, and the `options` array is not empty.

[record-2025-02-14_14.00.16.webm](https://github.com/user-attachments/assets/2b1b67e1-3087-45ce-b19c-043189b86458)

packages/jupyter-chat/src/components/input/use-chat-commands.tsx
```diff
@@ -0,0 +1,168 @@
...
+      }
+
+      const currentWord = getCurrentWord(input, cursorIndex);
+      if (!currentWord?.length) {
+        return;
+      }
+
+      let newCommands: ChatCommand[] = [];
+      for (const provider of providers) {
+        // TODO: optimize performance when this method is truly async
```
brichet, line 62: Maybe this optimization could be perform at the `provider` level.

For example, it could keep in memory the previous word and suggestions. If the new word is only additional letters from the previous one, it only filters the previous suggestions.

dlqqq: Agreed! Provider-level optimizations will be necessary for more complex command providers that need to make network requests frequently. However, what I meant by this comment was that we can also do some optimizations in our frontend:

- Use `Promise.all()` to start async `getChatCommands()` tasks in parallel.
- Show chat command suggestions as they resolve, instead of waiting for every provider to finish getting chat commands.

dlqqq: Opened a new issue for this: https://github.com/jupyterlab/jupyter-chat/issues/167

packages/jupyter-chat/src/components/input/use-chat-commands.tsx
dlqqq: Good thinking. I'll update the code & resolve this thread when I do so.

packages/jupyter-chat/src/chat-commands/types.ts
dlqqq: Agreed, we should keep the interface simple. However, the `currentWord` and `replaceCurrentWord` params will be replaced with `chatModel/inputModel` once we start on #163. Therefore, I think we are free to ignore this concern for now.

dlqqq: I've added #163 to the epic so we won't forget about this.

packages/jupyter-chat/src/components/input/use-chat-commands.tsx
dlqqq: I duplicated `open` because `chatCommands.menu.open` is more readable than `chatCommands.autocompleteProps.open`. I also found it helpful to distinguish the props being passed to the `Autocomplete` component (under `autocompleteProps`) from the state we are exposing to `ChatInput` (under `menu`).

We are still using a single source of truth (the `open` state variable), so I don't see a concern here. However, feel free to push changes to this branch if you think differently.

packages/jupyterlab-chat/src/factory.ts
dlqqq: Agreed! This is one of the follow-up items I called out. I'll open a subissue for this.

packages/jupyter-chat/src/components/input/use-chat-commands.tsx
dlqqq: Done!

python/jupyterlab-chat/src/chat-commands/slash-command-demo.ts
dlqqq: It would make more sense to provide a demo after the `IChatCommandProvider` API is finalized. This would also be a great first issue for new contributors, so I'll open a new issue to see if we can get interest.

python/jupyterlab-chat/src/chat-commands/registry.ts
dlqqq: Done. Please keep this conversation open for now so I will remember to open an issue.

dlqqq:
@brichet @jtpio Thank you both for the review. I've addressed your feedback and fixed the bugs that @brichet noticed.

I'll close the remaining threads once I finish opening issues. Unfortunately, I have something really urgent to attend to at work right now, so I'll finish that on Monday hopefully.

dlqqq:
Demo of latest branch state.

https://github.com/user-attachments/assets/1e4157e7-506c-4fe5-b21c-5ebfbb46a8da

python/jupyterlab-chat/src/chat-commands/registry.ts
brichet: > BTW, how do you comment a file directly? I've never seen that in a GitHub code review 👀

The `comment` icon on the top right corner

Screenshot from 2025-02-17 09-37-07

brichet: > However, I have a feeling we could simplify this repo to just 2 packages: jupyterlab-chat should provide all the tokens, and jupyterlab-chat-extension should provide all the implementations.

The `jupyterlab-chat` provides the modules for a chat based on shared documents, while `@jupyter/chat` provides mostly the front-end components. Indeed, it also includes some token that can be useful for any chat using this library.

For me the question is most about the separated `jupyterlab-chat` and `jupyterlab-chat-extension` that could be reduce to one package. This separation is mostly to follow the pattern used in jupyterlab.

Feel free to open an issue about it if you think that it deserves more discussions.

brichet (commented):
Thanks @dlqqq for updating the PR, I can't reproduce the issues anymore :+1: .

packages/jupyter-chat/src/components/input/use-chat-commands.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    const replaceCurrentWordClosure = (newWord: string) => {
+      replaceCurrentWord(input, cursorIndex, newWord, setInput);
+    };
+
+    chatCommandRegistry.handleChatCommand(
+      command,
+      currentWord,
+      replaceCurrentWordClosure
+    );
```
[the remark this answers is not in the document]
dlqqq: I agree! I'm just keeping this here to indicate that we should implement a way of getting & setting the current word in the `InputModel` (#163). Some commands may need to replace the current word with something dynamic (i.e. not statically definable in `ChatCommand.replaceWith`).

python/jupyterlab-chat/src/chat-commands/registry.ts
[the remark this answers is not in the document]
dlqqq: Done! 👍

brichet:
Thanks @dlqqq.
The failing test are related to the old autocompletion framework.
The best option would be to update them.
Otherwise we should at least skip them, and maybe add a fix to https://github.com/jupyterlab/jupyter-chat/issues/168 (or another issue).

python/jupyterlab-chat/src/chat-commands/registry.ts
dlqqq: New issue: #173.

dlqqq:
@brichet Since CI is already failing on `main`, can we go ahead and merge this while tracking the CI failure in a separate issue? Merging this now would make it a lot easier to work on #171.

python/jupyterlab-chat/src/chat-commands/slash-command-demo.ts
dlqqq: Removed the demo. Opened a new issue for the Jupyterlite demo: #178.

dlqqq:
Finally finished opening all of the follow-up issues raised by other reviewers and me! Thank you all for taking the time to review this & ask questions. 🤗

jtpio:
> Since CI is already failing on `main`, can we go ahead and merge this while tracking the CI failure in a separate issue?

Not sure it's the same failure. Based on @brichet's comment above (https://github.com/jupyterlab/jupyter-chat/pull/161#issuecomment-2663606332) it seems related to the previous autocompletion registry implementation:

image

Ideally we should be able to update these tests? If it's too much effort for now, maybe worth disabling / removing them for now, but track adding them back in an issue?

dlqqq:
@jtpio Thanks for the reminder! I've opened a new issue: #179

jtpio:
Nice thank! 

So let's remove or disable the outdated test code in this PR?

After https://github.com/jupyterlab/jupyter-chat/pull/148 the CI should normally be passing more consistently again (with skipped tests). And to lower maintenance burden it's nice to have a green check even though we know some tests have been disabled (it's easier to merge other PRs).

dlqqq:
@jtpio Good idea, I'll do that now.

dlqqq:
@jtpio  Done! 👍

brichet (approved):
Thanks @dlqqq and @jtpio

PR #162, opened 2025-02-11
Lite deployment in doc

### LINKED ISSUE #149: JupyterLite demo
### Problem

It would be useful to have a small demo of `jupyter-chat` running in JupyterLite, to be able to quickly get a feel of the current UX and functionalities it offers.

### Proposed Solution

Add a JupyterLite demo with the `jupyter-chat` extension, ideally built from source, so it reflects the latest state.

### Additional context

Currently `jupyterlab-chat` defines a list of dependencies that require a server to be functional:

https://github.com/jupyterlab/jupyter-chat/blob/8de473aa176a5fbdf8ff7165ff629606b5d36a74/python/jupyterlab-chat/pyproject.toml#L28-L34

Ideally it would be possible to try the chat and most of its frontend features directly in the browser, without any RTC involved. Or if there is RTC, then maybe over WebRTC?

brichet:
Fixes #149

Add a jupyterlite demo in the documentation, including a simple chat extension, to be able to test the features.
This extension does not allow to chat for real (with others users) since collaboration is not working in jupyterlite yet.


jtpio (approved):
Thanks!

PR #171, opened 2025-02-17
Add new `InputModel` class for managing input state

### LINKED ISSUE #163: Subissue: Lift chat input state to `ChatModel`
### Problem

While building #161, I noticed that I had to pass `input` and `setInput` from the `<ChatInput />` component to the chat command providers, since they need to read & write from the input. This results in a weird interface:

```ts
export interface IChatCommandProvider {
    id: string;
    ready?: Promise<void>;
    getChatCommands(partialInput: string): Promise<ChatCommand[]>;
    handleChatCommand(command: ChatCommand, partialInput: string, replacePartialInput: (newPartialInput: string) => void): Promise<void>;
}
```

If a `ChatModel` instance could provide methods for reading & writing to the input, the interface can then be simplified to:

```ts
export interface IChatCommandProvider {
    id: string;
    ready?: Promise<void>;
    getChatCommands(chatModel: IChatModel): Promise<ChatCommand[]>;
    handleChatCommand(command: ChatCommand, chatModel: IChatModel): Promise<void>;
}
```

### Proposed Solution

Lift chat input state up to the `ChatModel` instance. The `<ChatInput />` component should be updated accordingly.

### Additional context

Chat commands PR: #161

brichet:
This PR adds an `InputModel` class to handle input state.
This is proposed for:
- simplifying the use of the chats commands, that needs to get the input content, the current word and to be able to replace it.
- enhancing the messages editiing, by allowing adding and deleting attachments.

By default, the chat model includes an input model, for the "main" input of the chat panel. When editing a message, a new input model is created for this edition, and deleted when edition is over (cancelled or validated).

Fixes #163


dlqqq:
@brichet I think this PR will cause some conflicts with the chat commands PR in #161. I think it would be best to rebase this branch & fix conflicts after merging #161. I'll give this a review after.

dlqqq:
#148 should probably also be merged first to address any merge conflicts. I've approved that PR.

packages/jupyter-chat/src/components/input/use-chat-commands.tsx
```diff
@@ -101,34 +100,20 @@ export function useChatCommands(
...
-      replaceCurrentWord(input, cursorIndex, command.replaceWith, setInput);
-      return;
-    }
-
-    const replaceCurrentWordClosure = (newWord: string) => {
-      replaceCurrentWord(input, cursorIndex, newWord, setInput);
-    };
+    // When a command is selected, the input value is not updated to avoid conflict
+    // changes. It must be updated here, even if there is no replacement string.
+    inputModel.replaceCurrentWord(command.replaceWith ?? command.name);
```
brichet, line 114: This change is to avoid concurrency on the input update.
When a command is selected from the list, the `onInputChange()` callback of the `ChatInput` is called first with the value (name) of the command. Then the `onChange()` callback of the `Autocomplete` is called. This resulted to a replacement of the whole input.
To avoid this behavior, the `onInputChange()` does not update the input value when change come from a selected command (using the `reason` props, set to `reset` in that case), and the value is updated in the `onchange()`, whether it need to be replaced or not.

This seems to work smoothly, but maybe it deserves some more testing to ensure it did not break anything else.

cc. @dlqqq

packages/jupyter-chat/src/chat-commands/types.ts
```diff
@@ -67,7 +68,6 @@ export interface IChatCommandProvider {
    */
   handleChatCommand(
     command: ChatCommand,
-    currentWord: string,
-    replaceCurrentWord: (newWord: string) => void
+    inputModel: IInputModel
   ): Promise<void>;
```
dlqqq, line 72: Can you also pass the input model to `getChatCommands()` method above on line 60? The `currentWord: string` argument should be replaced with `inputModel: IInputModel`.

packages/jupyter-chat/src/components/chat-input.tsx
```diff
@@ -17,33 +18,27 @@ import React, { useEffect, useRef, useState } from 'react';
...
+import { IAttachment, Selection } from '../types';
 import { useChatCommands } from './input/use-chat-commands';
 import { IChatCommandRegistry } from '../chat-commands';
 
 const INPUT_BOX_CLASS = 'jp-chat-input-container';
 
 export function ChatInput(props: ChatInput.IProps): JSX.Element {
   const { documentManager, model } = props;
-  const [input, setInput] = useState<string>(props.value || '');
+  const [input, setInput] = useState<string>(model.value);
```
dlqqq, line 31: I'm hesitant to leave this line here because it creates 2 sources of truth about what the current state of the input is (`model.value` v.s. `input`). It seems like we should remove the `input` and `setInput` variables entirely. Can you explore making these changes?

packages/jupyter-chat/src/components/chat-input.tsx
```diff
@@ -269,11 +276,7 @@ export namespace ChatInput {
     /**
      * The chat model.
      */
-    model: IChatModel;
-    /**
-     * The initial value of the input (default to '')
-     */
-    value?: string;
+    model: IInputModel;
```
dlqqq, line 279: This is a nitpick, but can we prefer naming input models just `input` instead of `model` or `inputModel`? The word `model` doesn't convey anything useful. Preferring `input` produces code that is easier to follow:

```ts
input.value // value of the input
input.attachments // the input attachments
input.cursorIndex // index of the cursor in the input
```

Same goes for the arguments to the `IChatCommandProvider` methods.

dlqqq (changes requested):
@brichet Wow, thanks for working on this so quickly! All of the high-level changes look good. I left some feedback for you below, but still need more time to complete a full review.

My day tomorrow will be full of meetings, so I will likely need until the end of Thursday to finish reviewing this. Feel free to work on another issue using the Kanban board: https://github.com/orgs/jupyterlab/projects/9

packages/jupyter-chat/src/components/chat-input.tsx
brichet: As discussed privately, the state is used to update the react component.
`setInput()` should only be called when the input model value has changed, I don't think that it can lead to errors.

packages/jupyter-chat/src/chat-commands/types.ts
brichet: Done

packages/jupyter-chat/src/components/chat-input.tsx
dlqqq: I agree. I did 30 min of research into React rendering and it doesn't seem like there's any other way to trigger a re-render. This leads to a little duplicate state, but I think that's fine given that there's no clear alternative.

packages/jupyter-chat/src/components/chat-input.tsx
dlqqq: As discussed in a previous thread, we still need `input` as a local state variable to trigger React re-renders. Therefore this comment isn't applicable.

packages/jupyter-chat/src/components/input/use-chat-commands.tsx
dlqqq: @brichet Thank you for documenting this. I've tested it locally and it works well. The code highlighted here actually introduces a small regression, because we only want to replace the current word if `replaceWith` is set. If it's not set, then the handling of the command should be deferred to the command provider. I've pushed this change to your branch, and have verified this locally.

Note: I actually added that `if (reason === 'reset') { ... }` condition in the original PR, but removed it later because I found that it wasn't necessary. However, based on my local testing, this PR definitely needs that condition. I'm not sure why this bug only surfaced here in this PR. Leaving this as a note in case this breaks in the future.

dlqqq (approved):
@brichet This looks perfect, thank you for building this!

packages/jupyter-chat/src/components/input/use-chat-commands.tsx
brichet: > because we only want to replace the current word if replaceWith is set.

Let's keep your change, I don't have a strong opinion on it :+1:

packages/jupyter-chat/src/components/input/use-chat-commands.tsx
```diff
@@ -101,34 +100,23 @@ export function useChatCommands(
...
     if (!currentWord) {
       return;
     }
 
+    // if replaceWith is set, handle the command immediately
     if (command.replaceWith) {
-      replaceCurrentWord(input, cursorIndex, command.replaceWith, setInput);
+      inputModel.replaceCurrentWord(command.replaceWith);
       return;
     }
```
brichet, line 116: Should it be one or the other only (`replaceWith` or `handleChatCommand()`) ?
We could imagine replacing the word and performing an action. Removing the `return` statement allows it without counterpart, I think.

dlqqq: I do think that they should be mutually exclusive, since `replaceWith` can be implemented in `handleChatCommand()` if necessary. If a command provider only provides commands with `replaceWith` set, then the command provider is only providing completions. These commands don't need to be handled, so `handleChatCommand()` can be left blank (i.e. it just returns immediately).

If a command provider wants to replace the current word and perform an action, then they should define their custom logic in `handleChatCommand()`. For example:

```tsx
// additional logic before
inputModel.replaceCurrentWord(...)
// additional logic after
```

The current working model is that command providers **should** define `handleChatCommand()` in any case where they need to handle the command themselves, i.e. they aren't just providing completions. This is understandably unclear, since we only provide completions with emojis right now.

brichet:
Thanks for the review @dlqqq

PR #188, opened 2025-03-11
Move the CSS rules of chat commands in @jupyter/chat

brichet:
This PR moves the CSS rules of chat commands from `jupyterlab-chat` to `@jupyter/chat` (where they are defined).

`@jupyter/chat` can be used as a frontend extension that does not depends on `jupyterlab-chat`.
Therefore, the CSS rules should be defined in `@jupyter/chat`.


jtpio (approved):
Thanks!

jtpio:
Starting a new `0.8.1` patch release now.

PR #190, opened 2025-03-11
Mention users in messages (using @)

### LINKED ISSUE #184: Subissue: Allow users to be mentioned
### Problem

It must be possible to `@`-mention other users in the chat. Typing `@` should open the completions menu by default and show a list of users and their avatars.

### Proposed Solution

1. Use the chat commands framework to implement this capability.
2. Show user avatars in the chat commands menu.
3. (optional) Explore ways to show an inline toast for a mentioned user, e.g. `@Nicolas` should be highlighted, clickable, and in blue.
4. Add the user to a list of mentions.

### Additional context

Does there exist a method in the frontend to retrieve the current list of users from `IChatModel`? If not, that also needs to be implemented.

brichet:
This PR allows to mention users in messages.
Fixes #184 

[record-2025-03-20_10.55.07.webm](https://github.com/user-attachments/assets/df8da77d-bedf-4326-b5cc-aa21bf11480a)

### General workflow

- The autocomplete contains all users who have already written in the chat and all users who are currently connected to it. Each chat command displays the user avatar.
- Selecting a user from the chat commands updates a list of mentions in the input model, and adds `@username` in the input.
- When the message is sent, the mentioned users are attached to the message object.
- When the message is displayed (markdown renderer), the mentioned user string (`@username`) is replaced by a span to add style on it.

### Some details on the changes

- The `IChatCommandProvider` methods receive also the `ChatModel` in arguments. It's required to find the users of the chat.
- The `IUser` interface has now an optional attribute `mention_name` (string), used to find the mention in the message and to replace it with the span on the fly.
- When editing a message, the mention span is replaced back to the `mention_name`, for a better user experience.
- It seems difficult to catch when a mention is removed from the message, we should listen for changes in message and looks for `user_mention` in real time. Therefore, when the message is sent (or updated), the mention list of the input is checked to make sure that the message still contains the `mention_name`.


brichet:
The current implementation creates the user list every time the mention command is requested (typing `@` in the input).

It seems to be quite reactive with a few users, but may become problematic if a lot of users are involved/connected to the chat.
It might be possible to build and update the user list for each chat using signals.

The `MentionCommandProvider` could receive `app.serviceManager.events.stream` signal to build the list of users when a chat is created, and clean it when a chat is deleted.
Then we would have to listen for (1) the user list in the shared model (users who wrote in the chat or have been mentioned) and (2) the user list in the shared model awareness (users currently connected to chat) to update the users of each chat in `MentionCommandProvider`.

Maybe we should wait for existing issues before implementing this.

python/jupyterlab-chat/src/chat-commands/providers/user-mention.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+};
+
+class MentionCommandProvider implements IChatCommandProvider {
+  public id: string = 'jupyter-chat:mention-commands';
+
+  // regex used to test the current word
+  private _regex: RegExp = /^@[\w-]*:?/;
+
+  async getChatCommands(inputModel: IInputModel, chatModel: LabChatModel) {
```
dlqqq, line 35: Passing the entire `chatModel` to each command provider may be providing an excessive amount of scope. Ideally, chat command providers should only have control over the input, and should not be able to affect the entire chat state.

It seems like it may make more sense to replace the `chatModel: ChatModel` argument with a `chatContext: ChatContext` object, a custom object that provides read-only APIs to the chat, like `getUsers()`. That would also greatly simplify this method, since `MentionCommandProvider` could just call `chatContext.getUsers()` to get the list of users.

dlqqq (changes requested):
@brichet Thank you for working on this! This is a huge contribution.

I left some feedback below. Not sure how much of it we want to address in this PR. I think you're in a better position to judge what can / can't be addressed here.

BTW, does this allow typing `@Jupyternaut` when Jupyter AI is installed?

packages/jupyter-chat/src/utils.ts
```diff
@@ -45,3 +49,33 @@ export function getCellIndex(notebook: Notebook, cellId: string): number {
...
+ * @param user - the user mentioned.
+ */
+export function replaceSpanToMention(content: string, user: IUser): string {
+  if (!user.mention_name) {
+    return content;
+  }
+  const span = `<span class="${MENTION_CLASS}">${user.mention_name}</span>`;
+  const regex = new RegExp(span, 'g');
+  return content.replace(regex, user.mention_name);
+}
```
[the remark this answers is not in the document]
brichet: > 1. This may introduce a XSS vulnerability because it's unclear whether `user.mention_name` is sanitized before rendering. React might do this automatically, but if not, a malicious actor could exploit this by setting their username to `Good person <script src="https://evil.com/evil.js" />`.

Good catch, we should definitely address this one before moving forward

> I believe the ideal long-term solution is to extend the Chat Commands framework to allow replacing the current word with a HTML element within the chat box.

The chat box is currently a `textarea`, which does not allow us to insert HTML Elements. The current behavior is the same as in Zulip for example I think.
For Slack, the input seems to be an editable div containing HTML elements. We can probably update the input to use a library providing these features (Quill, Code mirror...) in a follow up PR.

python/jupyterlab-chat/src/chat-commands/providers/user-mention.ts
brichet: :+1: I agree with you, and maybe that read-only chat model could be passed to the `ChatInput`, that way it could also be used for the buttons in the toolbar registry.

packages/jupyter-chat/src/utils.ts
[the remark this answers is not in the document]
dlqqq: @brichet Thanks for the quick research! I opened a new issue to track this: #204.

python/jupyterlab-chat/src/chat-commands/providers/user-mention.ts
dlqqq: @brichet That sounds good! Do you think that this should be done in a follow-up PR, or should we address it here?

Either way sounds fine to me, but I think we should solve this before the next release. Otherwise we might have to migrate Jupyter AI v3 to use `chatModel`, only to change it later to use `chatContext`.

brichet: I'll update this PR

brichet:
> BTW, does this allow typing @Jupyternaut when Jupyter AI is installed?

Yes, Jupyternaut is connected to the chat as a regular user: https://github.com/jupyterlab/jupyter-ai/blob/a82eecc220a032b9bfbef9beb3da900a8bfd4e59/packages/jupyter-ai/jupyter_ai/extension.py#L272

brichet:
> This may introduce a XSS vulnerability because it's unclear whether user.mention_name is sanitized before rendering. React might do this automatically, but if not, a malicious actor could exploit this by setting their username to Good person <script src="https://evil.com/evil.js" />.

IIUC the default rendermime [registry](https://github.com/jupyterlab/jupyterlab/blob/8ce37a58199674b1a7a44e2f5cb059465fcef411/packages/rendermime-extension/src/index.ts#L138) has a sanitizer, which doesn't allow `script` tags : https://github.com/jupyterlab/jupyterlab/blob/8ce37a58199674b1a7a44e2f5cb059465fcef411/packages/apputils/src/sanitizer.ts#L490

The additional span is displayed using the markdown renderer, so it should be sanitized. I did a test and the `script` tag is indeed removed from the rendered message.
Do you think that it'senough to prevent XSS attack (at least in a regular installation with the sanitizer available) ?

packages/jupyter-chat/src/input-model.ts
```diff
@@ -3,19 +3,26 @@
...
+import { IAttachment, IUser } from './types';
 
 const WHITESPACE = new Set([' ', '\n', '\t']);
 
 /**
  * The chat input interface.
  */
-export interface IInputModel extends IDisposable {
+export interface IInputModel<T extends IChatContext = IChatContext>
+  extends IDisposable {
```
dlqqq, line 20: If we're only providing a single implementation of `IInputModel`, can this type be made non-generic? Generic types make the code more complex for multiple reasons:

1. A generic interface requires a user to inspect the interface & read through it to determine what the type argument `T` is specifying.
2. Type arguments often cannot be inferred, which encourages unsafe type casting (e.g. `as any`) as a quick fix.
3. A generic type argument `T` doesn't provide any insight into what `T` may possibly be. By contrast, a union like `type ChatContext = ChatContextA | ChatContextB` makes it clear that it can only be 1/2 of types, and [discriminators](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) can be used to narrow the type of an instance of `ChatContext`.

To address this, I recommend that we:

1. Explore if a single type can be defined for `ChatContext`,
2. And if that is not possible, prefer defining `ChatContext` as a union of subtypes instead of as a generic type argument.

packages/jupyterlab-chat-extension/src/chat-commands/providers/user-mention.ts
```diff
@@ -0,0 +1,98 @@
...
+  private _regex: RegExp = /^@[\w-]*:?/;
+
+  async getChatCommands(inputModel: IInputModel) {
+    this._users.clear();
+    const match = inputModel.currentWord?.match(this._regex)?.[0];
+    if (!match) {
+      return [];
+    }
+
+    const users = (inputModel.chatContext as LabChatContext).users;
```
dlqqq, line 41: This is an example of unsafe type casting encouraged by generic types. If `chatContext` were a union, this could be made type-safe by replacing this with:

```ts
// inputModel: IChatContext
if (inputModel.type !== 'lab') {
  return [];
}

// now inputModel is of type 'LabChatContext'
```

dlqqq (changes requested):
@brichet Love the new chat context! I thought it would be a second argument, but placing it under `InputModel` still makes sense & keeps the API smaller. 🤗 

I'm not able to give a full review today, but I was able to find an opportunity to simplify the code further. Let me know what you think! 👋

packages/jupyter-chat/src/model.ts
```diff
@@ -648,3 +673,32 @@ export namespace IChatModel {
...
+  constructor(options: { model: IChatModel }) {
+    this._model = options.model;
+  }
+
+  get name(): string {
+    return this._model.name;
+  }
+
+  protected _model: IChatModel;
+}
```
dlqqq, line 704: Forgot to add this to my review.

Just curious, what is the reasoning behind have a `ChatContext` and a `LabChatContext`? It may not be clear which one is the "main" implementation used in Jupyter Chat. Instead, it could be better to stick to one, and let developers provide their own implementation if they want to use another. What do you think?

dlqqq (changes requested):

dlqqq:
@brichet 

> Do you think that it'senough to prevent XSS attack (at least in a regular installation with the sanitizer available) ?

Yes, and thank you for doing that research! It's safe to replace `@`-mentions with `<span>` elements for now, but I think we should improve this (#204) before JAI v3.0.0.

packages/jupyter-chat/src/model.ts
brichet: The `ChatContext` is the default implementation of the context, while `LabChatContext` is extending it to use it in `jupyterlab-chat`.

packages/jupyterlab-chat/src/model.ts
```diff
@@ -349,3 +427,37 @@ export class LabChatModel
 
   private _user: IUser;
 }
+
+/**
+ * The chat context to be sent to the input model.
+ */
+export class LabChatContext extends ChatContext {
```
dlqqq, line 434: Do we need to provide both `LabChatContext` and `ChatContext`? There are other areas in this branch where we are unsafely casting `IChatContext` objects as `LabChatContext` to access the `users` property defined here. If we require this property, then it doesn't seem to make sense to have the `IChatContext` exclude that property. 

I recommend that the `users` property be added to the `IChatContext` interface to remove the type casts. I also suggest we merge `ChatContext` and `LabChatContext` together, since one interface will only require one implementation. Is this possible? If not, what is the reason for providing 2 implementations in 2 different packages?

dlqqq (changes requested):
Thank you for the updates @brichet! There are still some unsafe type casts being introduced by this branch, and I think we should continue exploring ways to keep the code safe. Left a comment below that clarifies one of my earlier comments.

dlqqq:
Documenting some usability issues I noticed while testing this branch. No need to address these in the same PR, just making a list so we can open an issue later.

- Unable to un-focus the chat input within the chat widget.
- Previously-connected users are listed in the commands menu when `@` is typed. Even if the same user is connecting to the chat repeatedly, the list of users grows and becomes a list of `@Anonymous-XYZ`, `@Anonymous-FooBar`, etc.

dlqqq:
@brichet I've opened a new PR targeting this branch, which adds some of the feedback that I had suggested in my review: https://github.com/brichet/jupyter-chat/pull/4

brichet:
> @brichet I've opened a new PR on your fork targeting this branch. This implements the feedback that I had suggested in my last review: [brichet#4](https://github.com/brichet/jupyter-chat/pull/4)

Thanks @dlqqq for opening it.
I give some comments here, for a better visibility.

With this "double" implementation of the `ChatContext` (in `@jupyter/chat` and `jupyterlab-chat`), I wanted to avoid:
- force a third party extension (extending the `AbstractChatModel`) to implement the `createChatContext()` when it does not need it
- force the implementation of `get users()` when extending the `IChatContext` if not necessary

This is why I found it simpler to provide a default `ChatContext` anyway, that can be extended only if necessary.

For example, the `AbstractChatModel` is extended in [jupyterlite-ai](https://github.com/jupyterlite/ai/blob/7e2c434c66fc316f1b247c4eb11624afee989663/src/chat-handler.ts#L39):
- (1) in the current state, there is no need for a `ChatContext`
- (2) if we add a `ChatContext` in future, the concept of `users` does not really make sense

-------

1. If you really think that we should avoid type casting (I don't have a strong opinion on it), do you think it is possible to make `createChatContext()` optional in the `AbstractChatModel`, and `get user()` optional in the `IChatContext` ?


2. An other option could be to safely cast type, maybe something like 
```ts
if (inputModel.chatContext implements ILabChatContext):
```

dlqqq:
@brichet But doesn't it still make sense for JupyterLite to implement this interface? I understand that there is no RTC in JupyterLite since there is no backend. However, JupyterLite can either just return an empty array or an array with one item describing the current user. For example:

```ts
class LiteChatContext extends AbstractChatContext implements IChatContext {
   /**
   * The list of users who have connected to this chat. 
   */
  get users(): IUser[] {
    return []; // or, return a one-item array like [this.user]
  }
}
```

Even though it doesn't seem like JupyterLite needs to implement `IChatContext` right now, it may require this in the future. If not, we can make `createChatContext()` optional later on, before v1.0.0.

PR #191, opened 2025-03-11
Fix the trailing hyphen in completion without description

brichet:
This PR deletes the trailing hyphen in the autocomplete suggestion (chat commands), if the description is empty.

Example:


brichet:
I merged it, this is not a big change and can easily be reverted if necessary.

PR #194, opened 2025-03-20
Fix duplication after message edition

### LINKED ISSUE #183: Mesage edition display both messages
## Description

Modifying a message result to both messages displayed in the chat.
The previous message is not saved in the shared document, and disappears when reloading the page.

https://github.com/user-attachments/assets/762b1012-22ef-4361-92b1-6778f496418b

## Reproduce

Edit a message from the chat panel.

## Expected behavior

The previous message to disappear.

brichet:
Fixes #183 
 
[record-2025-03-20_14.17.08.webm](https://github.com/user-attachments/assets/1e48815e-22cc-42a7-82c0-191cd5578597)

When a message is edited, the changes received from the `YChat` are an insertion and a deletion (in the order).
Inserting a message in the `LabChatModel` is asynchronous, because it needs to wait for the model to be ready (which means to have an ID). This is required for the initialization of the message list.

Before deleting the message, we now wait for the messages to be inserted, to get the correct index of the message to delete.

### DIFF
--- a/packages/jupyterlab-chat/src/model.ts
+++ b/packages/jupyterlab-chat/src/model.ts
@@ -133,10 +133,13 @@ export class LabChatModel extends ChatModel implements DocumentRegistry.IModel {
     // nothing to do
   }
 
-  messagesInserted(index: number, messages: IChatMessage[]): void {
+  async messagesInserted(
+    index: number,
+    messages: IChatMessage[]
+  ): Promise<void> {
     // Ensure the chat has an ID before inserting the messages, to properly catch the
     // unread messages (the last read message is saved using the chat ID).
-    this._ready.promise.then(() => {
+    return this._ready.promise.then(() => {
       super.messagesInserted(index, messages);
     });
   }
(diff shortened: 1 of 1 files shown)


dlqqq (approved):
Thank you! This change makes sense and looks good to me.

It seems like there's a couple of issues that can occur when deleting messages from the list. Managing the index of each message in the array of chat messages introduces a fair amount of complexity throughout the codebase.

I wonder if it would be simpler to never delete elements from the list of chat messages, and completely ignore indexes. Message deletion can be expressed as a message edit. For example, deleting a message is kind of the same as just marking it as deleted and showing "message deleted" in the UI.

If we follow this pattern throughout Jupyter Chat, this brings two major benefits:

1. We always know exactly what index each message is.
2. We don't have to ever update that index or re-calculate indices.

This is also essentially free in terms of memory use. In Yjs, items are never deleted, but simply marked as deleted in the data model. Deleting a message doesn't really save any memory anyways. If this change can simplify the codebase, I think we should take this opportunity.

@brichet What do you think?

brichet:
Thanks @dlqqq for the review, I opened #203 to track the message deletion.

PR #195, opened 2025-03-20
Add a clearMessages() method in ChatModel

### LINKED ISSUE #169: Clear method to clear messages
### Problem

Currently consumers of `IChatModel` need to use the following method to clear all messages:

https://github.com/jupyterlab/jupyter-chat/blob/73512c16a6c602b81f873ecf9a3c20cb73cefd94/packages/jupyter-chat/src/model.ts#L157

For example:

```py
this.messagesDeleted(0, this.messages.length);
```

### Proposed Solution

Maybe it could be more convenient if a `clear` method was exposed?

### Additional context

One objection would be that adding a new method increases the API surface.

brichet:
Fix #169


jtpio (approved):
Thanks!

Issue #165, opened 2025-02-14
Move `python/jupyterlab-chat` to `packages/jupyterlab-chat-extension`?


dlqqq:
### Problem

It's a little strange how `python/jupyterlab-chat`:

- Shares a similar name to `packages/jupyterlab-chat`,
- Refers to a NPM package of a different name (`jupyterlab-chat-extension`), and
- Is under a different directory than the other extensions in this monorepo.

### Proposed Solution

Move `python/jupyterlab-chat` to `packages/jupyterlab-chat-extension`.

Jupyter AI's monorepo structure allows for Python server extensions under `packages/` as well. It uses a custom Jupyter Releaser configuration and Lerna scripts to build & release the packages.

### Additional context

It would be nice to solve this issue to make it easier for new contributors to ramp up. However, this is a low priority issue given that it has no user impact. 

@brichet Can you share more details about why the repo has a separate `python/` folder, and if you think this issue makes sense?

brichet:
Agree, maybe we should do like in `jupyter-collaboration`, keep only the python part in the `python` directory, and move `jupyterlab-chat-extension` in packages.

PR #196, opened 2025-03-20
Move jupyterlab-chat-extension to packages directory

brichet:
This PR moves the `jupyterlab-chat-extension` from the `python` directory to the `packages` directory.

This has been discussed in https://github.com/jupyterlab/jupyter-chat/issues/165

It also use a python script instead of a bash script for dev installation, for a better compatibility with Windows systems (see https://github.com/jupyterlab/jupyter-chat/pull/88#pullrequestreview-2363075125).


brichet:
Thanks for the review @dlqqq

PR #197, opened 2025-03-21
Abstract ChatModel class

### LINKED ISSUE #157: Abstact `ChatModel` in `@jupyter/chat`
> I got a bit lost by the inheritance structure while reviewing this. We have `IChatModel`, then we have `ChatModel` implementing that interface (defined in `@jupyter/chat`), and then we have `LabChatModel` extending `ChatModel` in `jupyterlab-chat`.

> As a reviewer, I was confused about why the Jupyter Chat project provides two implementations of `IChatModel`.

> It seems like `ChatModel` in `@jupyter/chat` should be an abstract base class, and `ChatModel` should be renamed to `AbstractChatModel`. This would help make the inheritance structure more obvious. What are your thoughts on this?

_Originally posted by @dlqqq in https://github.com/jupyterlab/jupyter-chat/pull/148#discussion_r1943716519_

brichet:
The `ChatModel` class in `@jupyter/chat` package is not intended to be instantiated but extended, since it does not provide the `sendMessage()` implementation.

This PR abstracts this class for clarification.

Fixes #157

### DIFF
--- a/packages/jupyter-chat/src/model.ts
+++ b/packages/jupyter-chat/src/model.ts
@@ -180,11 +180,11 @@ export interface IChatModel extends IDisposable {
  * The class inheriting from it must implement at least:
  * - sendMessage(message: INewMessage)
  */
-export abstract class ChatModel implements IChatModel {
+export abstract class AbstractChatModel implements IChatModel {
   /**
    * Create a new chat model.
    */
-  constructor(options: ChatModel.IOptions = {}) {
+  constructor(options: IChatModel.IOptions = {}) {
     if (options.id) {
       this.id = options.id;
     }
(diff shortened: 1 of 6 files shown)


packages/jupyter-chat/src/model.ts
```diff
@@ -169,11 +169,12 @@ export interface IChatModel extends IDisposable {
...
- * The default chat model implementation.
- * It is not able to send or update a message by itself, since it depends on the
- * chosen technology.
+ * An abstract implementation of IChatModel.
+ *
+ * The class inheriting from it must implement at least:
+ * - sendMessage(message: INewMessage)
  */
-export class ChatModel implements IChatModel {
+export abstract class ChatModel implements IChatModel {
```
dlqqq, line 177: Can this also be renamed to `AbstractChatModel` for clarity?

dlqqq (changes requested):
@brichet Thanks! Just one tiny comment.

packages/jupyter-chat/src/model.ts
brichet: Thanks, I updated it.

Issue #172, opened 2025-02-17
Use a `JupyterlabContext` React context


brichet:
Would it be helpful to build some kind of `JupyterlabContext` React context which provides access to all the APIs provided by other JupyterLab extensions?

For example, I was thinking that a component could call something like:

```ts
const { commandRegistry } = useJupyterlabContext();
```

and similarly so to access other JupyterLab singletons. This would simplify the frontend code by not requiring us to pass every singleton as a prop to every child component. This is also equivalent in terms of performance, as these singleton objects are not changed at runtime.

_Originally posted by @dlqqq in https://github.com/jupyterlab/jupyter-chat/pull/148#discussion_r1947140049_
            

This context could probably includes most of the properties of the [`ChatBodyProps`](https://github.com/jupyterlab/jupyter-chat/blob/73512c16a6c602b81f873ecf9a3c20cb73cefd94/packages/jupyter-chat/src/components/chat.tsx#L111), but also the `activeCellManager` and the `selectionWatcher` that are currently in the model.

dlqqq:
I have a fair amount of experience with React contexts & TypeScript. @brichet Do you want me to own this issue?

dlqqq:
In `chat.tsx`, we want to pass these 4 props to a new React context provider to allow children to access them via context hooks:


```tsx
    /**
     * The rendermime registry.
     */
    rmRegistry: IRenderMimeRegistry;
    /**
     * The document manager.
     */
    documentManager?: IDocumentManager;
    /**
     * Chat command registry.
     */
    chatCommandRegistry?: IChatCommandRegistry;
    /**
     * Attachment opener registry.
     */
    attachmentOpenerRegistry?: IAttachmentOpenerRegistry;
```

There are also 2 extra unused props: 


```tsx
    /**
     * Autocompletion registry.
     */
    autocompletionRegistry?: IAutocompletionRegistry;
    /**
     * Autocompletion name.
     */
    autocompletionName?: string;
```

These unused props are part of the old autocompletion framework (implemented in #57) and can be removed. These are part of another issue: #168.

brichet:
> There are also 2 extra unused props:
> 
>     /**
>      * Autocompletion registry.
>      */
>     autocompletionRegistry?: IAutocompletionRegistry;
>     /**
>      * Autocompletion name.
>      */
>     autocompletionName?: string;

Removed in https://github.com/jupyterlab/jupyter-chat/pull/187

PR #198, opened 2025-03-25
Input toolbar registry

### LINKED ISSUE #170: Extensible toolbar
### Problem

Currently it looks like `jupyter-chat` does not provide the ability to stop message streaming. Or a way to provide streaming functionality from another extension.

### Proposed Solution

JupyterAI v2 seems to be supporting this via https://github.com/jupyterlab/jupyter-ai/pull/1022

Image

This would likely require porting some of the work originally done in https://github.com/jupyterlab/jupyter-ai/pull/1022.

Probably `jupyter-chat` should only focus on providing the "Stop" button, and a way to customize behavior on stop via plugins.

### Additional context

Currently Jupyter AI v3 (in dev) does not seem to be supporting that:

Image

brichet:
Fixes #170


packages/jupyter-chat/src/components/chat.tsx
```diff
@@ -134,6 +121,10 @@ export namespace Chat {
      * Attachment opener registry.
      */
     attachmentOpenerRegistry?: IAttachmentOpenerRegistry;
+    /**
+     * The input toolbar registry
+     */
+    inputToolbarRegistry?: IInputToolbarRegistry;
```
brichet, line 127: Reference to https://github.com/jupyterlab/jupyter-chat/issues/172, this property could also be added to a `ReactContext`.

python/jupyterlab-chat/src/index.ts
```diff
@@ -791,13 +799,32 @@ const selectionWatcher: JupyterFrontEndPlugin<ISelectionWatcher> = {
...
+const inputToolbarFactory: JupyterFrontEndPlugin<IInputToolbarRegistryFactory> =
+  {
+    id: pluginIds.inputToolbarFactory,
+    description: 'The input toolbar registry plugin.',
+    autoStart: true,
+    provides: IInputToolbarRegistryFactory,
+    activate: (app: JupyterFrontEnd): IInputToolbarRegistryFactory => {
+      return {
+        create() {
+          return InputToolbarRegistry.defaultToolbarRegistry();
```
jtpio, line 814: Not sure how much work it would be, but could the JupyterLab `IToolbarWidgetRegistry` be reused?

In the case of the send / stop buttons, maybe instead of two distinct buttons it could be encapsulated as one widget, which renders differently based on some condition?

brichet: > Not sure how much work it would be, but could the JupyterLab IToolbarWidgetRegistry be reused?

Thanks for the comment, I'll take a look at it, it may be worth using indeed.

> In the case of the send / stop buttons, maybe instead of two distinct buttons it could be encapsulated as one widget, which renders differently based on some condition?

In the example above, there's no others buttons, but if there were, I suppose we'd want to hide them too. For example an *attachments* button, that would not be necessary during the bot response.

brichet: As far as I understand, `IToolbarWidgetRegistry` is designed to work with lumino `Widget`. The chat widget is full react (embedded in a `ReactWidget`), so I don't really know how it can be compatible without a big refactoring of the chat itself.

jtpio: ok, then let's think about that later. Thanks for checking!

packages/jupyter-chat/src/components/input/toolbar-registry.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  /**
+   * Add a toolbar item.
+   */
+  addItem(name: string, item: InputToolbarRegistry.IToolbarItem): void;
+
+  /**
+   * Remove a toolbar item.
+   */
+  removeItem(name: string): void;
```
jtpio, line 37: If this method is not used yet, maybe we can remove it from the interface, and add it later if needed?

jtpio (approved):
Thanks @brichet, looks good overall!

Just left a small comment, but it would be nice to merge and release afterwards so the change can be integrated into downstreams like Jupyter AI.

brichet:
Thanks @jtpio for the review

PR #200, opened 2025-04-01
Fix the attachments on message edition and on python Message model

brichet:
Follow up #148.

The python `Message` model was missing the `attachments` property, which led to error when trying to cast some message dict to `Message` objects:

```bash
TypeError: Message.__init__() got an unexpected keyword argument 'attachments'
```

Should fixes issues [2] and [4] in https://github.com/jupyterlab/jupyter-ai/issues/1293


jtpio (approved):
Thanks!

PR #207, opened 2025-04-11
Fix _version.py file version bump

brichet:
Follow up #196 to fix the version of the `_version.py` file in python extension when bumping to a specific value (e.g `0.10.0`)


brichet:
Merging it to try a release.

Issue #202, opened 2025-04-02
FR: Modular message footer API


dlqqq:
### Problem

Upstream issue in Jupyter AI: https://github.com/jupyterlab/jupyter-ai/issues/1297

As discussed there, #198 adds a new feature that allows Jupyter Chat to replicate the "Stop streaming" capability in Jupyter AI v2. Thank you all for working on this! 🤗 

However, I think that we should also explore other ways of allowing users to stop streaming messages. Users should be able to send messages regardless of whether an AI is streaming a reply to this user. The UI in Jupyter AI v2 made sense because it was just one user w/ one AI back then, but it doesn't scale well to many users w/ many AIs.

Here, I propose building a **modular message footer API** to:

1. Allow Jupyter AI to provide a "stop streaming" button at the bottom of each chat message in the center, and
2. Allow other labextensions to add something to the bottom of each chat message, without clobbering Jupyter AI's "Stop streaming" button.

The reason I seek 2) is because other labextensions may want to add message footers as well. [This feature was added in Jupyter AI 2.x](https://github.com/jupyterlab/jupyter-ai/pull/942) for other applications to provide custom UI elements to each message. For example, some developers have used this API to render thumbs up / thumbs down buttons to allow users to "rate" an AI response.

### Proposed design

1. Jupyter Chat optionally requests the `ChatMessageFooter` token. When other plugins provide this, they have to provide a React component that renders their footer, and pick a location to render. The React component accepts the current message & position as props. The location can be "left", "right", or "center".

```tsx
export const ChatMessageFooter = new Token<IChatFactory>(
  'jupyterlab-chat:ChatMessageFooter'
);

export interface ChatMessageFooterProps {
  message: Message
  position: 'left' | 'center' | 'right'
}

export interface ChatMesageFooter {
  component: React.FC<ChatMessageFooterProps>
  position: 'left' | 'center' | 'right'
}
```

2. Jupyter Chat accepts all message footers and builds an object like this:

```ts
type Footers = {
  Left?: ChatMessageFooter
  Center?: ChatMessageFooter
  Right?: ChatMessageFooter
}
```

If 2 plugins provide the same footer at the same location, we can just pick 1 arbitrarily. The idea is that if one labextension is occupying the 'center' position, another labextension can render their footer on the 'left' or 'right' position. Therefore these conflicts should not happen to begin with.

3. Jupyter Chat then renders the `footers` object above in a flexbox, like this:

```tsx
[...]

PR #210, opened 2025-04-16
Add a message footer registry, to allow extensions to provides their footers

brichet:
This PR adds a message footer registry, allowing to add 3 footers components to each message, at left, center and right position.

If the registry is provided to the chat (which is the case with `jupyterlab-chat` extension), the functional components will be called with the chat model and the message as argument. This allows to display a specific footer depending on the message itself and to react to some model signals.

For example in `jupyter-ai`, it will allow to add a stop streaming button in the footer messages. When Jupyternaut is answering, the `writers` list of the model is updated and a signal is emitted. With #208, this `writers` list will also contains the message ID that is edited.
The stop streaming button could listen to the `writers` signal and display a stop streaming button if (1) the bot is writing and (2) the message including this footer is the one that is edited.

Fixes https://github.com/jupyterlab/jupyter-chat/issues/202
Related to https://github.com/jupyterlab/jupyter-ai/issues/1297


packages/jupyter-chat/src/components/messages/footer.tsx
brichet, line 1: This module is currently the only one in `components/messages` but I'd like to split the current `chat-messages.tsx` into several modules (`message.tsx`, `header.tsx`, `navigation.tsx`, `toolbar.tsx`...) for a better readability.

packages/jupyterlab-chat/src/factory.ts
```diff
@@ -130,6 +135,7 @@ export namespace ChatWidgetFactory {
     chatCommandRegistry?: IChatCommandRegistry;
     attachmentOpenerRegistry?: IAttachmentOpenerRegistry;
     inputToolbarFactory?: IInputToolbarRegistryFactory;
+    chatFooterRegistry?: IChatFooterRegistry;
```
brichet, line 138: We should probably reuse the `IChatBodyProps` from `@jupyter/chat` instead of listing all of the properties here.

brichet (commented):
Notes on code refactoring while working on this PR.

packages/jupyter-chat/src/footers/types.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/**
+ * The footers returned by the registry.
+ */
+export type Footers = {
+  left?: IChatMessageFooter;
+  center?: IChatMessageFooter;
+  right?: IChatMessageFooter;
+};
```
dlqqq, line 32: Only one comment: I think we can improve these names to be more clear. Right now, "Footer" refers to both the group and a individual within that group.

A footer is everything at the bottom of some parent text, so it's more accurate to refer to the group of 3 as "the footer", and refer to each individual as a "footer section". I also think we can drop the prefix "IChat" from these names, since "message" almost always refers to "chat message" in this project, and interfaces don't need to be named explicitly.

```suggestion
/**
 * The props sent passed to each `MessageFooterSection` React component.
 */
export type MessageFooterSectionProps = {
  model: IChatModel;
  message: IChatMessage;
}

/**
 * A message footer section which can be added to the footer registry.
 */
export type MessageFooterSection = {
  component: React.FC<MessageFooterSectionProps>;
  position: 'left' | 'center' | 'right';
}

/**
 * The message footer returned by the registry, composed of 'left', 'center',
 * and 'right' sections.
 */
export type MessageFooter = {
  left?: MessageFooterSection;
  center?: MessageFooterSection;
  right?: MessageFooterSection;
};
```

packages/jupyter-chat/src/footers/registry.ts
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Get the footer from the registry.
+   */
+  get(): Footers;
+  /**
+   * Add a message footer.
+   * If several extension add footers, only the last one will be displayed.
+   */
+  add(footer: IChatMessageFooter): void;
+}
```
dlqqq, line 22: Similarly:

```suggestion
export interface IMessageFooterRegistry {
  /**
   * Get the message footer.
   */
  getFooter(): MessageFooter;
  /**
   * Add a message footer section.
   * If multiple labextensions add a section in the same region, only
   * the last one will be displayed.
   */
  addSection(section: MessageFooterSection): void;
}
```

dlqqq (commented):
🤗 ❤️ Thank you for working on this! Code looks great. Just left a couple of comments that suggest more consistent names for the new things introduced by this PR.

Otherwise this is good to go! 🚀

brichet:
Thanks for the review @dlqqq, I updated the names according to your suggestions.

packages/jupyter-chat/src/footers/types.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import { IChatModel } from '../model';
+import { IChatMessage } from '../types';
+
+/**
+ * The props sent passed to each `MessageFooterSection` React component.
+ */
+export type MessageFooterSectionProps = {
+  model: IChatModel;
```
brichet, line 13: Maybe we should send only the `IChatContext` here, in which we could add some `ChatModel` signals. That way the component could be reactive the changes in the model without accessing the whole methods of the model.

packages/jupyter-chat/src/components/messages/footer.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * The chat footer component properties.
+ */
+export interface IMessageFootersProps extends MessageFooterSectionProps {
+  /**
+   * The chat footer registry.
+   */
+  registry: IMessageFooterRegistry;
+}
```
dlqqq, line 21: One more naming suggestion:

```suggestion
export interface MessageFooterProps extends MessageFooterSectionProps {
  /**
   * The chat footer registry.
   */
  registry: IMessageFooterRegistry;
}
```

dlqqq (approved):
@brichet The new names are great, thank you! I also left another minor naming suggestion for you, up above. PR looks fantastic otherwise. 🤗 ❤️

packages/jupyter-chat/src/components/messages/footer.tsx
brichet: It does not fit with the naming convention, an interface must start with `I`.
Let's keep it that way for now, we'll change it to type later if it make more sense.

PR #225, opened 2025-06-06
Jupyter chat package reorganization

brichet:
This PR is a reorganization of the `@jupyter-chat` package, it is only maintenance.
There should be no breaking change and no changes in the UI.

The main changes are:
- split of the `chat-messages` module into several modules in `components/messages` (e.g. _footer.tsx_, _header.tsx_, _message.tsx_, _navigation.tsx_...)
- centralize all the registers in a dedicated directory: _attachment-openers_, _chat-commands_ and _footers_


dlqqq (approved):
This looks great, thank you!

PR #226, opened 2025-06-10
Code highlight in message

### LINKED ISSUE #222: Code attached from the cell has no code highlighting in the chat windows
## Description

Code attached from the cell has no code highlighting in the chat windows.

## Context

See https://github.com/jupyterlab/jupyter-ai/issues/1356 for context

brichet:
This PR add code highlighting when  it is attached from a cell or a selection.

image

The language is added to the markdown code block when it is available.

- Code attached from a cell
  Get the language from the `defaultKernelLanguage` of the Notebook metadata. If not defined, there is no syntax highlighting.

- Code attached from a selection
  Retrieve the `DocumentWidget` mimetype from its editor [model](https://github.com/jupyterlab/jupyterlab/blob/1f18828225e5e16c569dcb810eb4c6137cd33d9b/packages/codeeditor/src/editor.ts#L144), and convert it to a language using the [`EditorLanguageRegistry`](https://github.com/jupyterlab/jupyterlab/blob/1f18828225e5e16c569dcb810eb4c6137cd33d9b/packages/codemirror/src/language.ts#L61). 
  If the mimetype cannot be converted to a language,there is no syntax highlighting.

### Context 

Fixes #222
Related to https://github.com/jupyterlab/jupyter-ai/issues/1356

### DIFF
--- a/packages/jupyter-chat/src/components/input/buttons/send-button.tsx
+++ b/packages/jupyter-chat/src/components/input/buttons/send-button.tsx
@@ -109,7 +109,7 @@ export function SendButton(
     // Run all chat command providers
     await chatCommandRegistry?.onSubmit(model);
 
-    let language = '';
+    let language: string | undefined;
     if (selectionWatcher?.selection) {
       // Append the selected text if exists.
       source = selectionWatcher.selection.text;
@@ -118,13 +118,13 @@ export function SendButton(
       // Append the active cell content if exists.
       const content = activeCellManager.getContent(false);
       source = content!.source;
-      language = content?.language || '';
+      language = content?.language;
     }
     let content = model.value;
     if (source) {
       content += `
 
-\`\`\`${language}
+\`\`\`${language ?? ''}
 ${source}
 \`\`\`
 `;
--- a/packages/jupyter-chat/src/components/input/buttons/send-button.tsx
+++ b/packages/jupyter-chat/src/components/input/buttons/send-button.tsx
@@ -109,18 +109,21 @@ export function SendButton(
     // Run all chat command providers
     await chatCommandRegistry?.onSubmit(model);
 
+    let language = '';
     if (selectionWatcher?.selection) {
       // Append the selected text if exists.
       source = selectionWatcher.selection.text;
     } else if (activeCellManager?.available) {
       // Append the active cell content if exists.
-      source = activeCellManager.getContent(false)!.source;
+      const content = activeCellManager.getContent(false);
+      source = content!.source;
+      language = content?.language || '';
     }
     let content = model.value;
     if (source) {
       content += `
 
-\`\`\`
+\`\`\`${language}
 ${source}
 \`\`\`
 `;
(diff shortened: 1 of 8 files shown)


packages/jupyter-chat/src/selection-watcher.ts
```diff
@@ -44,6 +52,10 @@ export namespace SelectionWatcher {
      * The ID of the document widget in which the selection was made.
      */
     widgetId: string;
+    /**
+     * The language of the selection.
+     */
+    language: string;
```
andrii-i, line 58: If we to follow suggestion, `language` could / should be declared as optional.

packages/jupyter-chat/src/selection-watcher.ts
```diff
@@ -207,13 +225,15 @@ function getTextSelection(
     [start, end] = [end, start];
   }
 
+  const language = await languages.getLanguage(editor?.model.mimeType);
```
andrii-i, line 228: Note `getLanguage` return type is `Promise<IEditorLanguage | null>` so it also returns "not set" in an explicit manner but using `null`.

packages/jupyter-chat/src/selection-watcher.ts
```diff
@@ -207,13 +225,15 @@ function getTextSelection(
...
   return {
     ...selectionObj,
     start,
     end,
     text,
     numLines: text.split('\n').length,
     widgetId: widget.id,
+    language: language?.name || '',
```
andrii-i, line 236: If we to follow suggestion, if `language` is `null` it does not need to be returned at all.

andrii-i (commented):
Thank you for working on this @brichet. I tested this PR locally and it works well. It also improves user experience as missing code highlighting in the snippets attached to the chat messages was noticeable. 

Please find small suggestion on making typing of `language` property more explicit below.

packages/jupyter-chat/src/components/input/buttons/send-button.tsx
```diff
@@ -100,19 +100,22 @@ export function SendButton(
...
+      const content = activeCellManager.getContent(false);
+      source = content!.source;
+      language = content?.language;
     }
     let content = model.value;
     if (source) {
       content += `
 
-\`\`\`
+\`\`\`${language ?? ''}
```
dlqqq, line 118: There are potentially 3 different lists of languages:

1. From file editor models: `(await languages.getLanguage(editor?.model.mimeType))?.name;`
2. From notebook models: `this._notebookTracker.currentWidget?.model?.defaultKernelLanguage`
3. The list of language identifiers recognized by the Rendermime plugin that renders these code blocks
 
We should try to find source code references for all 3. If this lists do not exactly match, that means there are edge cases in which this does not work. We don't need to handle them all, but we should probably open an issue that includes all 3 lists for future developers to fix these edge cases.

packages/jupyter-chat/src/selection-watcher.ts
```diff
@@ -207,13 +225,18 @@ function getTextSelection(
     [start, end] = [end, start];
   }
 
+  const language = (await languages.getLanguage(editor?.model.mimeType))?.name;
```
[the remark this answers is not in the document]
brichet: Probably, I don't think the content is read to find determine the language. It may come from https://github.com/jupyterlab/jupyterlab/blob/main/packages/codemirror/src/mimetype.ts.

PR #233, opened 2025-06-18
Clean up dependencies in the Binder environment

jtpio:
Checking if we still need these dependencies to be listed in the Binder environment.

Especially since they appear to be outdated, and already defined here?

https://github.com/jupyterlab/jupyter-chat/blob/825371829006c86814375c76220a5abc688a296b/python/jupyterlab-chat/pyproject.toml#L29-L34


dlqqq (approved):
Looks awesome, thank you!

PR #235, opened 2025-06-18
Allow users to be mentioned via Python API

dlqqq:
## Description

- Addresses https://github.com/jupyterlab/jupyter-ai/issues/1365
- Updates the `YChat.add_message()` method to parse for `@`-mentions using the same logic implemented in the frontend. 
- `@`-mentioned usernames are added to the `mentions` field of the new message.
- `YChat.mention_name` has been updated to be a computed property that will always be defined, ensuring type safety.
- Adds unit test coverage for this change.

### DIFF
--- a/python/jupyterlab-chat/jupyterlab_chat/models.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/models.py
@@ -75,11 +75,14 @@ class NewMessage:
 class User(JupyterUser):
     """ Object representing a user """
 
-    mention_name: Optional[str] = None
-    """ The string to use as mention in chat """
-
     bot: Optional[bool] = None
     """ Boolean identifying if user is a bot """
+
+    @property
+    def mention_name(self) -> str:
+        name: str = self.display_name or self.name or self.username
+        name = name.replace(" ", "-")
+        return name
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


ellisonbg:
Nice, will have a look.

python/jupyterlab-chat/jupyterlab_chat/models.py
```diff
@@ -75,12 +75,15 @@ class NewMessage:
...
 
-    mention_name: Optional[str] = None
-    """ The string to use as mention in chat """
-
     bot: Optional[bool] = None
     """ Boolean identifying if user is a bot """
 
+    @property
+    def mention_name(self) -> str:
+        name: str = self.display_name or self.name or self.username
```
JGuinegagne, line 83: sanity-check: any risk that they are all nullish? as in should we add a fallback here?

python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -132,9 +133,19 @@ def add_message(self, new_message: NewMessage) -> str:
...
             time=timestamp,
-            id=uid
+            id=uid,
         )
 
+        # find all mentioned users and add them as message mentions
+        mention_pattern = re.compile("@([\w-]+):?")
+        mentioned_names: Set[str] = set(re.findall(mention_pattern, message.body))
+        users = self.get_users()
+        mentioned_usernames = []
```
JGuinegagne, line 143: sanity-check: is there a risk that this list goes > 10 and that `len(users.items())` goes > 10?

If so, consider using a set here.

dlqqq: Also good to check. This should be safe because we are iterating through the `users` dictionary to get all users, and checking to see if their mention name was in the set of mentions.

So since we check against the *set* of mentions, it doesn't matter if a user was mentioned 1 or 100 times; they will only be added to the `mentioned_usernames` list once. 👍

PR #242, opened 2025-06-25
Fix v0.14.0 bugs

### LINKED ISSUE #241: Minor bugs in Jupyter Chat v0.14.0 w/ JAI v3
## Description

This issue describes some minor bugs I had noticed w/ the 0.14.0 release when integrating it with Jupyter AI v3 (https://github.com/jupyterlab/jupyter-ai/pull/1386).

1. Personas cannot send new messages because the `YChat.add_message()` method raises an exception claiming `mention_name` is not a recognized argument to `User`.

2. This error log sometimes pops up in the browser console. When this happens, pressing "Enter" creates a new line instead of sending the message.
Image

3. Command providers cannot change the input before sending the message.

4. The frontend sets `user.mention_name = @<mention-name>`, but the backend sets `user.mention_name = <mention-name>` (without the `@` prefix). Furthermore, the `mention_name` property is optional in the frontend, despite always being computable. The data models must be consistent across the frontend & backend.

5. Finally, `inputModel.currentWord` should not terminate at ' ' space characters if they are escaped. This allows for commands to accept arguments that include spaces, e.g. file paths.

## Reproduce

Can be replicated by doing a developer installation from this branch: https://github.com/jupyterlab/jupyter-ai/pull/1386

dlqqq:
## Description

- Closes #241

This PR fixes each of the small bugs reported in #241. The commit history addresses each in the order they appear in #241.

The most notable changes are:

1. The `user.mention_name` property has been made consistent across the frontend & backend. It is now computed using the same procedure and is always defined in the `LabChatModel`.

2. The `inputModel.currentWord` property has been modified to allow for spaces escaped with a preceding backslash `\` in the current word. This allows commands to contain spaces, allowing `@file` in JAI to accept file paths that contain spaces.

### DIFF
--- a/packages/jupyter-chat/src/types.ts
+++ b/packages/jupyter-chat/src/types.ts
@@ -14,9 +14,14 @@ export interface IUser {
   color?: string;
   avatar_url?: string;
   /**
-   * The string to use to mention a user in the chat.
+   * The string to use to mention a user in the chat. This is computed via the
+   * following procedure:
+   *
+   * 1. Let `mention_name = user.display_name || user.name || user.username`.
+   *
+   * 2. Replace each ' ' character with '-' in `mention_name`.
    */
-  mention_name?: string;
+  mention_name: string;
   /**
    * Boolean identifying if user is a bot.
    */
--- a/docs/jupyter-chat-example/src/index.ts
+++ b/docs/jupyter-chat-example/src/index.ts
@@ -56,7 +56,7 @@ class MyChatModel extends AbstractChatModel {
       id: newMessage.id ?? UUID.uuid4(),
       type: 'msg',
       time: Date.now() / 1000,
-      sender: { username: 'me' },
+      sender: { username: 'me', mention_name: 'me' },
       attachments: this.input.attachments
     };
     this.messageAdded(message);
--- a/packages/jupyter-chat/src/components/input/chat-input.tsx
+++ b/packages/jupyter-chat/src/components/input/chat-input.tsx
@@ -159,7 +159,7 @@ export function ChatInput(props: ChatInput.IProps): JSX.Element {
     ) {
       // Run all command providers
       await props.chatCommandRegistry?.onSubmit(model);
-      model.send(input);
+      model.send(model.value);
       event.stopPropagation();
       event.preventDefault();
     }
--- a/packages/jupyter-chat/src/components/input/use-chat-commands.tsx
+++ b/packages/jupyter-chat/src/components/input/use-chat-commands.tsx
@@ -37,7 +37,9 @@ export function useChatCommands(
   // whether an option is highlighted in the chat commands menu
   const [highlighted, setHighlighted] = useState(false);
 
-  // whether the chat commands menu is open
+  // whether the chat commands menu is open.
+  // NOTE: every `setOpen(false)` call should be followed by a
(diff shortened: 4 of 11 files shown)


brichet:
@dlqqq thanks for working on this.

> 1. The `user.mention_name` property has been made consistent across the frontend & backend. It is now computed using the same procedure and is always defined in the `LabChatModel`.

It may be a naive question, but I wonder what would be the drawback to just build this `mention_name` every time the `setUser` method is called ([frontend](https://github.com/jupyterlab/jupyter-chat/blob/5bc764b508b9392147a2a5c2ef18156c0c0bee02/packages/jupyterlab-chat/src/ychat.ts#L161) and [backend](https://github.com/jupyterlab/jupyter-chat/blob/5bc764b508b9392147a2a5c2ef18156c0c0bee02/python/jupyterlab-chat/jupyterlab_chat/ychat.py#L101)). That would ensure that every user has the property and that the same pattern is used.

And and we want backward compatibility, we can probably add it on [first load](https://github.com/jupyterlab/jupyter-chat/blob/5bc764b508b9392147a2a5c2ef18156c0c0bee02/python/jupyterlab-chat/jupyterlab_chat/ychat.py#L261).

dlqqq:
@brichet The benefit of adding a `LabChatUser` with a `mention_name` computed property is that it allows getting a `mention_name` from any object with `display_name`, `name`, `username` fields.

This is important in `ChatContext.users`, which reads from the awareness states of each user. These dictionaries do not contain the `mention_name` attribute and are set by the docprovider extension from `jupyter_collaboration`, hence why a custom class is necessary.

This probably can be simplified, but the changes should be safe as-is since `LabChatUser` is only used in `LabChatModel`.

dlqqq:
@andrii-i  BTW, this PR can be tested by:

- Dev installing this Jupyter Chat branch first, then

- Dev installing this Jupyter AI v3 branch after: https://github.com/jupyterlab/jupyter-ai/pull/1386

I would recommend doing this in a new environment to avoid breaking an existing one.

andrii-i (approved):
Looks and works well, improves user experience. Thank you @dlqqq.

I've tested this separately and with Jupyter AI v3, mentions and commands/filenames with spaces work well in different scenarios (couldn't break it).

dlqqq:
@andrii-i Thank you!

Our team is working really hard on Jupyter AI v3 right now & needs this change to use the `main` branch. I'm going to merge this & publish a v0.15 release right now. This should be safe since we're still on v0, and everything that depends on v0 should be using strict version ceilings.

@jtpio @brichet If our changes / releases ever cause an issue, please let us know. We're happy to take responsibility for our PRs & fix any issues they may cause.

packages/jupyter-chat/src/types.ts
```diff
@@ -14,9 +14,14 @@ export interface IUser {
...
-   * The string to use to mention a user in the chat.
+   * The string to use to mention a user in the chat. This is computed via the
+   * following procedure:
+   *
+   * 1. Let `mention_name = user.display_name || user.name || user.username`.
+   *
+   * 2. Replace each ' ' character with '-' in `mention_name`.
    */
-  mention_name?: string;
+  mention_name: string;
```
jtpio, line 24: Could this have been kept optional?

PR #246, opened 2025-06-26
Make `user.mention_name` optional in frontend

### LINKED ISSUE #245: Make `mention_name` optional
## Description

https://github.com/jupyterlab/jupyter-chat/pull/242 changed `mention_name` to be required, while it used to be optional in 0.14.0:

https://github.com/jupyterlab/jupyter-chat/blob/e6b486baab62e4708c4506ab2d9fd38fd172fba6/packages/jupyter-chat/src/types.ts#L24

Unfortunately didn't have the chance to review the PR before it got merged, but left a comment in https://github.com/jupyterlab/jupyter-chat/pull/242#pullrequestreview-2960949863.

There are use cases of the chat where `mention_name` is not required.

## Reproduce

Use `@jupyter/chat` 0.15.0 in a project in a downstream project.

## Expected behavior

`mention_name` should stay optional.

## Context

Jupyter Chat 0.15.0.

dlqqq:
## Description

- Closes #245.

Makes `user.mention_name` optional in the frontend. See discussion in #245 for context.

### DIFF
--- a/packages/jupyterlab-chat-extension/src/chat-commands/providers/user-mention.tsx
+++ b/packages/jupyterlab-chat-extension/src/chat-commands/providers/user-mention.tsx
@@ -126,6 +126,8 @@ class MentionCommandProvider implements IChatCommandProvider {
    */
   _getUsers(inputModel: IInputModel): Map<string, Private.CommandUser> {
     const users = new Map();
+    // chatContext should be of type `LabChatContext`, so `users` should be of
+    // type `LabChatUser[]`, where `mention_name` is always defined.
     const userList = inputModel.chatContext.users;
     userList.forEach(user => {
       users.set(user.mention_name, {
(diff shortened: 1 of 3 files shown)


packages/jupyterlab-chat-extension/src/chat-commands/providers/user-mention.tsx
```diff
@@ -126,8 +126,18 @@ class MentionCommandProvider implements IChatCommandProvider {
...
     userList.forEach(user => {
+      if (!user.mention_name) {
+        console.error(
+          `No 'mention_name' property for user '${user.username}'. ` +
+            'This user is being omitted from ' +
+            "'MentionCommandProvider._getUsers()'."
+        );
+        return;
+      }
       users.set(user.mention_name, {
```
brichet, line 141:
```suggestion
    const userList = inputModel.chatContext.users as LabChatUser[];
    userList.forEach(user => {
      users.set(user.mention_name, {
```

Since the `chatContext` is a `LabChatContext`, the user list should always be a `LabChatUser[]` type.

dlqqq: This is intentional even though I agree this is safe right now. The root issue is that `IChatContext.users` is not the same as `LabChatContext.users`, so it's possible that `LabChatContext.users` returns a user without a `mention_name`. In this case, we should log an error instead of ignoring the user silently.

PR #262, opened 2025-08-08
Adding multi chats panel in `@jupyter/chat` package

### LINKED ISSUE #261: Add a multi chats side panel in @jupyter/chat package
### Problem

`@jupyter/chat` package doesn't provide a multi-chats panel out of the box.

`jupyterlab-chat` package [does](https://github.com/jupyterlab/jupyter-chat/blob/98898f104681b85808a477af2b91d10a3a49745c/packages/jupyterlab-chat/src/widget.tsx#L101), its side panel allows to add/open/close chats, and to move them to the main area.

### Proposed Solution

We could make the `ChatPanel` from `jupyterlab-chat` generic enough to be provided by `@jupyter/chat`, and reused in ``jupyterlab-chat`.

### Additional context

Related to:
- https://github.com/jupyterlite/ai/issues/111
- https://github.com/jupyterlite/ai/pull/131

nakul-py:
Adding multi chats pannel to `@jypyter/chat` package and removing it from `jyputerlab-chat`. Now it can be easily imported and usable in `juputerlab-chat` and `jupyterlite/ai`.

- Closes #261


brichet:
Thanks @nakul-py for working on this.

I feel like this PR is not ready for review yet (e.g. we should not need the content manager in that chat, IMO).
Don't you think it should be converted to draft?

nakul-py:
@brichet These things have done: 
-  Defined `ready` promise in the `AbstractChatModel`.
-  Defined `IInputToolbarRegistry` in `@jupyter/chat`.
- Added the Rename chat button.

Will you test it now and suggest some improvements?

nakul-py:
> Thanks @nakul-py for updating it. FYI I won't be able to review it until September.

Ok @brichet i am waiting...

nakul-py:
please update snapshots

packages/jupyter-chat/src/multiChatPanel.tsx
```diff
@@ -0,0 +1,557 @@
...
+      const oldPath = path;
+      const newPath = PathExt.join(this.defaultDirectory, newName);
+
+      const ext = '.chat';
+      if (!newName.endsWith(ext)) {
+        newName += ext;
+      }
+
+      const contentsManager = new ContentsManager();
+      await contentsManager.rename(oldPath, newPath);
```
brichet, line 249: There shouldn't be any reference to path in this widget.
The notion of file/path is relevant only for `jupyterlab-chat`, which uses collaborative documents as chat.

packages/jupyter-chat/src/multiChatPanel.tsx
```diff
@@ -0,0 +1,557 @@
...
+ */
+export namespace ChatSection {
+  /**
+   * Options to build a chat section.
+   */
+  export interface IOptions extends Panel.IOptions {
+    widget: ChatWidget;
+    path: string;
+    defaultDirectory: string;
+    openChat: (path: string) => void;
```
brichet, line 521: This is not used in the `ChatSection` I think, only in the panel.

packages/jupyter-chat/src/model.ts
```diff
@@ -328,6 +337,18 @@ export abstract class AbstractChatModel implements IChatModel {
     localStorage.setItem(`@jupyter/chat:${this._id}`, JSON.stringify(storage));
   }
 
+  /**
+   * Promise that resolves when the model is ready.
+   */
+  readonly ready: Promise<void>;
```
brichet, line 343:
```suggestion
  get ready(): Promise<void> {
    return this._readyDelegate.promise;
  }
```
We could use a getter function.

packages/jupyter-chat/src/model.ts
```diff
@@ -677,6 +698,7 @@ export abstract class AbstractChatModel implements IChatModel {
   private _id: string | undefined;
   private _name: string = '';
   private _config: IConfig;
+  private _readyDelegate: PromiseDelegate<void>;
```
brichet, line 701:
```suggestion
  private _readyDelegate = new PromiseDelegate<void>();
```
Not sure it needs to be set in the constructor.

packages/jupyter-chat/src/multiChatPanel.tsx
brichet, line 1: This widget should probably be located in _src/widgets_ for consistency.

brichet (commented):
Thanks again @nakul-py for working on this, this is not an easy change.

This is not a full review.
My main comment is that there shouldn't be any notion of file/path in the generic multi-chat panel (`@jupyter/chat`) and section, this is only relevant for the `jupyterlab-chat` widget.

That said, I don't have a clear view of the best way to create a multi-chat panel in `jupyterlab-chat`:
- we can create a `MultiChatPanel` with functions in options to create, rename, open (your current implementation)
- we can have a new class in `jupyterlab-chat` that inherits from `MultiChatPanel`, and overrides some of its methods

If you prefer, I can also take it from now, and push in your PR.

packages/jupyter-chat/src/model.ts
nakul-py: Yes, the constructor doesn’t need to do anything for it.

packages/jupyter-chat/src/multiChatPanel.tsx
nakul-py: `openChat` is used by [moveToMain](https://github.com/jupyterlab/jupyter-chat/pull/262/files#diff-3e9698aab94f5d242f5c6c35b83b166e7211705d0d8a7a1620485e21e8f0fcacR383) button.
But `moveToMain`  is  actually not used anywhere inside `ChatSection`, so we should remove it

brichet:
> > If you prefer, I can also take it from now, and push in your PR.
> 
> Ok @brichet you can also work on this pr and now renaming, opening, closing and movetomain functions all are working fine.

Thanks @nakul-py, I'll probably push some commits in the next days.

nakul-py:
> I'll implement it in a follow up PR, to avoid having too many changes in this one.

Makes sense

PR #268, opened 2025-09-04
Fix chat focus behavior to allow copy/paste from messages

### LINKED ISSUE #267: Remove the `onclick` handler on `ChatWidget`?
Currently clicking on the chat brings the focus to the input.

This seems to be coming from here in the code base:

https://github.com/jupyterlab/jupyter-chat/blob/98898f104681b85808a477af2b91d10a3a49745c/packages/jupyter-chat/src/widgets/chat-widget.tsx#L43

This leads to the following behavior, making it difficult to copy past content from the chat:

https://github.com/user-attachments/assets/9f464112-e343-406b-bcd5-af83696b256b

Should this be kept, or removed?

nakul-py:
Previously, clicking anywhere in the chat panel forced focus onto the input (`this.node.onclick = () => this.model.input.focus();`). This made it difficult to select or copy text from the chat history, since every click redirected focus.

Now we check if the user has selected text input loses focus, click on `button`,` interactive elements`, `<details>`, `<summary>` and else work like before

https://github.com/user-attachments/assets/4f9b995e-01f2-4b29-be33-ed6597782000

Closes #267

### DIFF
--- a/packages/jupyter-chat/src/widgets/chat-widget.tsx
+++ b/packages/jupyter-chat/src/widgets/chat-widget.tsx
@@ -43,15 +43,26 @@ export class ChatWidget extends ReactWidget {
     this.id = `jupyter-chat::widget::${options.model.name}`;
     this.node.addEventListener('click', (event: MouseEvent) => {
       const target = event.target as HTMLElement;
-      const selection = window.getSelection();
-      const hasTextSelected =
-        selection &&
-        selection.toString().length > 0 &&
-        target.closest(`.${MESSAGE_CLASS}`);
-
-      if (!hasTextSelected && !target.closest(`.${MESSAGE_CLASS}`)) {
-        this.model.input.focus();
+      if (
+        target.closest('button, a, input, textarea, select, summary, details')
+      ) {
+        return;
       }
+
+      const message = target.closest(`.${MESSAGE_CLASS}`);
+      if (message) {
+        const selection = window.getSelection();
+
+        if (
+          selection &&
+          selection.toString().trim() !== '' &&
+          message.contains(selection.anchorNode)
+        ) {
+          return;
+        }
+      }
+
+      this.model.input.focus();
     });
   }
 
--- a/packages/jupyter-chat/src/widgets/chat-widget.tsx
+++ b/packages/jupyter-chat/src/widgets/chat-widget.tsx
@@ -11,7 +11,7 @@ import React from 'react';
 import { Message } from '@lumino/messaging';
 import { Drag } from '@lumino/dragdrop';
 
-import { Chat, IInputToolbarRegistry } from '../components';
+import { Chat, IInputToolbarRegistry, MESSAGE_CLASS } from '../components';
 import { chatIcon } from '../icons';
 import { IChatModel } from '../model';
 import {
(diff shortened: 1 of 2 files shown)


packages/jupyter-chat/src/widgets/chat-widget.tsx
```diff
@@ -40,7 +41,12 @@ export class ChatWidget extends ReactWidget {
 
     this._chatOptions = options;
     this.id = `jupyter-chat::widget::${options.model.name}`;
-    this.node.onclick = () => this.model.input.focus();
+    this.node.addEventListener('click', (event: MouseEvent) => {
+      const target = event.target as HTMLElement;
+      if (!target.closest(`.${MESSAGE_CLASS}`)) {
```
brichet, line 46: Perhaps we could check if some text is selected in the message, and avoid focusing on the input if some text is selected.

brichet (commented):
Thanks @nakul-py for working on this.

I have some comment below, let me know if it sound reasonable.

packages/jupyter-chat/src/widgets/chat-widget.tsx
brichet: The idea was to focus the input when clicking anywhere on the chat (including messages), except if this click result in a selection in the message.

Maybe something like the following (to be tested):

```ts
  const message = target.closest(`.${MESSAGE_CLASS}`);
  if (message) {
    const selection = window.getSelection();

    // Do not focus the input if the clicked message has selected text
    if (
      selection &&
      message.contains(selection.anchorNode) &&
      selection.toString().trim() !== ''
    ) {
      return;
    }
  }
  this.model.input.focus();
```

jtpio: > The idea was to focus the input when clicking anywhere on the chat (including messages), except if this click result in a selection in the message.

Don't we also want to allow for clicking on buttons that may be inlined in the chat directly or in `<details>` sections, without forcing the focus on the text area?

jtpio:
Wondering if we really need to force the focus to the chat area at all?

As a user, it can be frustrating if the application forces the focus to be somewhere else than where we want it to be.

packages/jupyter-chat/src/widgets/chat-widget.tsx
```diff
@@ -40,7 +41,29 @@ export class ChatWidget extends ReactWidget {
...
     this._chatOptions = options;
     this.id = `jupyter-chat::widget::${options.model.name}`;
-    this.node.onclick = () => this.model.input.focus();
+    this.node.addEventListener('click', (event: MouseEvent) => {
+      const target = event.target as HTMLElement;
+      if (
+        target.closest('button, a, input, textarea, select, summary, details')
+      ) {
+        return;
+      }
```
brichet, line 49: Thinking again about it, a more generic way could be to check if the current active element is on the chat, instead of checking if the target isl focusable element. Something like:

```suggestion
      if (this.node.contains(document.activeElement)) {
        return;
      }
```

That way we ensure that the input get focussed (and the tracker updates the current chat) only if the click occurs in a non-focusable element.

brichet: Thanks @nakul-py for updating it quickly.

Issue #279, opened 2025-09-23
Drag and drop on message edition


brichet:
## Description

Drag and drop to attach files does work when editing a message.

The file is attached to the main input instead of the editing input.

https://github.com/user-attachments/assets/db697362-9997-4924-a9ff-36dd74a56500

## Reproduce

Write a message, edit it and try to drag and drop a file in the edition input.

## Expected behavior

The file to be attached to the correct input.

PR #282, opened 2025-09-25
Add support for drag and drop while editing message.

nakul-py:
Adding support for drag and drop files while editing messages in chat input.

https://github.com/user-attachments/assets/508513a6-9c78-457b-b821-2556aea34f14

- Addresses #279


nakul-py:
> It seems that it does not handle several editions, neither dropping the file to the main input while editing.

Gonna fixing it :). Thanks for Attention!

nakul-py:
@brichet Any suggestion to make code changes more robust.

PR #284, opened 2025-09-26
Fix the main input height

brichet:
Follow up #278, to fix the input toolbar height.

- BEFORE

  [before.webm](https://github.com/user-attachments/assets/0f0f3859-ac4e-4a2f-a3f6-570aca398d14) 

- AFTER
- 
  [after.webm](https://github.com/user-attachments/assets/0f0d2da4-6208-40aa-bea7-0e15f7c7e7aa)


packages/jupyter-chat/style/chat.css
```diff
@@ -194,7 +194,7 @@
 }
 
 .jp-chat-navigation-bottom {
-  bottom: 100px;
+  bottom: 120px;
```
brichet, line 197: We should find a way to position this bottom navigation relatively to the input toolbar, instead of having a fixed size.

jtpio (approved):
Thanks!

PR #304, opened 2025-11-03
Lift the typing indicator out of the input component

### LINKED ISSUE #301: Lift typing indicator out of `<ChatInput />` component
### Problem

This is a follow-up issue tracking @brichet's feedback in #298:

> Is there any benefit of including the writing indicator component in the input component, instead of having it as sibling ? The input component is also used when editing a message, for example, and we won't use the writing indicator in this context.

brichet:
Fixes #301

This PR move the typing indicator out of the input component. This is mostly for convenience, there is no rational reason to bring the typing indicator when using the input component.
The UI does not change, the typing indicator is still displayed below the input.

### DIFF
--- a/packages/jupyter-chat/src/components/chat.tsx
+++ b/packages/jupyter-chat/src/components/chat.tsx
@@ -9,7 +9,7 @@ import ArrowBackIcon from '@mui/icons-material/ArrowBack';
 import SettingsIcon from '@mui/icons-material/Settings';
 import { IconButton } from '@mui/material';
 import { Box } from '@mui/system';
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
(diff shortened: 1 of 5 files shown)


ellisonbg:
This is a bit adjacent, but I recently saw the typing indicator for myself (was very confusing - I thought someone else was typing)

brichet:
> was very confusing - I thought someone else was typing

Sure, one might think that someone is logged into their session :smile: 

We assume that the user is defined as soon as the `serviceManager` is ready, but perhaps this assumption is not always true. https://github.com/jupyterlab/jupyter-chat/blob/5b4040e83374470375eb090683f4850f9df7dd2d/packages/jupyterlab-chat-extension/src/index.ts#L350-L352

PR #324, opened 2025-12-08
Making files dragable and dropable from TabBar.

nakul-py:
This PR makes files dragable and dropable from TabBar into the chat.

https://github.com/user-attachments/assets/fcc5cc34-226b-4694-ae2e-cc1a3f2a951e

This enhancement is done with the help of @brichet ❤️

💡 JupyterLite preview: https://jupyter-chat--324.org.readthedocs.build/en/324/lite/lab/index.html


packages/jupyter-chat/src/widgets/chat-widget.tsx
```diff
@@ -321,6 +328,41 @@ export class ChatWidget extends ReactWidget {
     }
   }
 
+  /**
+   * Process dropped tabBar files
+   */
+  private _processTabDrop(event: Drag.Event) {
+    const factory = event.mimeData.getData(TABBAR_FILE_MIME) as () => any;
```
brichet:
```suggestion
    const factory = event.mimeData.getData(TABBAR_FILE_MIME) as () => Widget;
```

The factory should always return return a `Widget`.

packages/jupyter-chat/src/widgets/chat-widget.tsx
```diff
@@ -321,6 +328,41 @@ export class ChatWidget extends ReactWidget {
...
+      console.warn('Factory did not return a widget');
+      return;
+    }
+
+    const path =
+      typeof widget.context?.path === 'string' ? widget.context.path : null;
+    if (!path) {
+      console.warn('Widget has no path');
+      return;
+    }
```
brichet:
```suggestion
    if (!widget || !(widget instanceof DocumentWidget) {
      console.warn('No file associated to the element');
      return;
    }

    const path = widget.context.path;
    if (!path) {
      console.warn('Widget has no path');
      return;
    }
```

Let's check for the Widget type, if it is not a `DocumentWidget`, no need to carry on.

packages/jupyter-chat/src/widgets/chat-widget.tsx
```diff
@@ -321,6 +328,41 @@ export class ChatWidget extends ReactWidget {
...
+
+    const mimetype =
+      widget.context?.model?.mimeType || 'application/octet-stream';
```
brichet:
```suggestion
      widget.context.model?.mimeType || 'application/octet-stream';
```

packages/jupyter-chat/src/widgets/chat-widget.tsx
nakul-py: Done 🎉

packages/jupyter-chat/src/widgets/chat-widget.tsx
nakul-py: Done 🎉

brichet:
Thanks @nakul-py, looks like the dependencies are missing from *package.json* file (and *yarn.lock*).
We should probably add the following dependency in *packages/jupyter-chat/package.json*:
```json
...
"@jupyterlab/docregistry": "^4.2.0",
...
"@lumino/widgets": "^2.0.0",
```

Then you'll need to run `jlpm install`, maybe test that it builds, and push the changes if everything is fine.

Issue #303, opened 2025-11-03
Follow up UX improvments


brichet:
#298 brings changes on UI/UX in the chat widget, but it also removed some existing features, that may require some more discussion.

Here is a non exhaustive list that can be updated:
- [x] **Edit** button, allowing you to edit a message after it has been sent. Some discussion about it [here](https://github.com/jupyterlab/jupyter-chat/pull/298#discussion_r2469011685) and [here](https://github.com/jupyterlab/jupyter-chat/pull/298#discussion_r2470309883)
  
  <img width="100" alt="Image" src="https://github.com/user-attachments/assets/bfa7e5e6-9f83-413d-b9ca-26cbcae841e5" />

- [ ] **Send with** dropdown, that was useful to add cell content or text selection in the message
  
  <img width="150" alt="Image" src="https://github.com/user-attachments/assets/19c7843f-e8a6-4cb3-a939-44d403809c52" />

- [x] **Message deleted** indicator. Before [#298](https://github.com/jupyterlab/jupyter-chat/pull/298), the message header was kept when a message had been deleted, to display it.

  <img width="176" height="51" alt="Image" src="https://github.com/user-attachments/assets/e831003e-c01c-46f8-abe3-446e6b5b012a" />

- [x] **Tooltip** with `Shift + Enter` action: send message or add new line, depending on the settings
- [x] Reuse the `TooltippedButton` for the toolbar buttons

PR #326, opened 2025-12-08
Fix the send button tooltip

brichet:
This PR fixes the tooltip on the send button.

~~This PR restore the helper text containing the shortcut to send a message.~~

Part of #303

### DIFF
--- a/packages/jupyter-chat/src/components/input/buttons/send-button.tsx
+++ b/packages/jupyter-chat/src/components/input/buttons/send-button.tsx
@@ -48,6 +48,9 @@ export function SendButton(
     };
     model.configChanged.connect(configChanged);
 
+    // Initialize the tooltip.
+    configChanged(model, model.config);
+
     return () => {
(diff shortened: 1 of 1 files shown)


brichet:
> Otherwise this could be a setting, to let administrators choose if they want to show it to their users?

It may be a bit overkill to add a settings for it. If we go in that direction, let remove the option to set it up, and keep one way only to send message and one to add a new line (this is not my favorite option, though).

> But maybe we can show this information in a different way, like a tooltip on the send button for example.

To make sure we're talking about the same thing, I was thinking of something like what we have in notebook buttons:

Screenshot from 2025-12-11 10-16-07

We could have a tooltip like `Send message (Enter)` or `Send message (Shift+Enter)`, and that could help deducting how to add a new line.

brichet:
Actually the tooltip was already there, but not initialized.

I updated the PR to revert the previous changes and fix tooltip initialization.

brichet:
Merging it, it is only a small bug fix.

PR #329, opened 2025-12-10
Exporting function ChatSelect

nakul-py:
Exporting function ChatSelect so  we can use it in `@jupyterlite/ai` to get list of chats.


brichet:
I'm closing this PR as the `ChatSelect` has been removed in #343

PR #334, opened 2025-12-13
Fixing Attachment opening in input bug

### LINKED ISSUE #333: Attachment doesn't open from input
## Description

When attaching a file to a message (drag and drop), the attachment is displayed in the input.
Clicking on it should open the file, but nothing happen.

Clicking on an attachment in the message after it has been sent works, though.

https://github.com/user-attachments/assets/1fbdbffc-1faa-4d3e-a0af-b63f0805bb03

## Reproduce

Attach a file to the message by dropping it in the input area.
Click on the file name, nothing happen.

## Expected behavior

The file to open.

## Context

It use to work, I don't when it has started to fail.

nakul-py:
Changing the key from a dynamically generated UUID to the stable `attachment.value` (the file path)

https://github.com/user-attachments/assets/655fbe42-d306-4bec-9b7e-8e4003c2b9b4

- Closes #333 

cc @brichet @jtpio 
- [x] Ready for review
- [x] PR has one of the labels: documentation, bug, enhancement, feature, maintenance

📚 Documentation preview: https://jupyter-chat--334.org.readthedocs.build/en/334/
💡 JupyterLite preview: https://jupyter-chat--334.org.readthedocs.build/en/334/lite/lab/index.html

### DIFF
--- a/packages/jupyter-chat/style/input.css
+++ b/packages/jupyter-chat/style/input.css
@@ -45,3 +45,7 @@
 .jp-chat-input-toolbar .jp-chat-send-include-opener {
   padding: 4px 0;
 }
+
+.jp-chat-input-container:focus-within > div:first-of-type {
+  border-color: var(--jp-brand-color1);
+}


packages/jupyter-chat/src/components/attachments.tsx
```diff
@@ -66,7 +65,7 @@ export function AttachmentPreviewList(props: AttachmentsProps): JSX.Element {
     >
       {props.attachments.map(attachment => (
         <AttachmentPreview
-          key={`${PathExt.basename(attachment.value)}-${UUID.uuid4()}`}
+          key={attachment.value}
```
brichet: Thanks @nakul-py for working on this.

This key has been added in https://github.com/jupyterlab/jupyter-chat/pull/248 to avoid duplicated key.

We should probably be careful by removing it.

After some tests, the issue is because of https://github.com/jupyterlab/jupyter-chat/blob/fbcdf8d3ee8d53e048d3b604c9e5feedf660f9aa/packages/jupyter-chat/src/components/input/chat-input.tsx#L217

Every time we click on the chat, the input get the focus, which update the state variable above. Therefore
&rightarrow; the react component (`ChatInput`) is reloaded
&rightarrow; this includes a reload of the attachments
&rightarrow; since the key is dynamic, the component is updated and the clicked attachment component is unmounted

Another option could be to remove this state and handle the border in the CSS.

brichet (approved):
Thanks @nakul-py

PR #351, opened 2026-02-07
Fix chat textbox stealing focus during text selection

veeceey:
## Summary
Fixes issue where the chat textbox aggressively steals focus, preventing users from selecting and copying text from AI responses.

This addresses [jupyterlab/jupyter-ai#1493](https://github.com/jupyterlab/jupyter-ai/issues/1493).

## Problem
Previously, clicking anywhere in the chat widget would focus the input textbox, even when users were trying to select text from responses. The click event handler checked for text selection, but the check happened at the wrong time - by the time the `click` event fired (after `mouseup`), the selection state wasn't properly captured.

## Solution
This PR implements a more robust focus management approach:

1. **Track mouse movement**: Detect drag gestures by comparing mousedown and click coordinates
2. **Delayed selection check**: Use `setTimeout(0)` to allow the selection to settle before checking
3. **Preserve UX**: Only prevent focus when user is actively selecting text in message areas

## Changes
- Added mousedown event listener to track initial mouse position
- Enhanced click event handler to detect selection attempts via mouse movement (>5px threshold)
- Added deferred selection check using setTimeout for message areas
- Preserved immediate focus behavior for clicks outside message areas

## Test Plan
**Manual Testing:**
1. Open Jupyter AI chat interface
2. Send a message and wait for AI response containing code or text
3. Click and drag to select partial text from the response
4. Release mouse button - selection should remain, focus should NOT move to input
5. Click in an empty area outside messages - focus should move to input (existing behavior)
6. Click directly on input - should focus normally (existing behavior)

**Expected Behavior:**
- ✅ Users can select and copy partial text from AI responses
- ✅ Input still gets focus when clicking in non-message areas
- ✅ Tab navigation to input still works
- ✅ No regression in existing focus management

## Related Issues
- Fixes jupyterlab/jupyter-ai#1493

---
*This fix improves the UX for users who need to copy partial responses from the AI, which is a common workflow when using code suggestions.*

### DIFF
--- a/packages/jupyter-chat/src/widgets/chat-widget.tsx
+++ b/packages/jupyter-chat/src/widgets/chat-widget.tsx
@@ -47,25 +47,54 @@ export class ChatWidget extends ReactWidget {
 
     this._chatOptions = options;
     this.id = `jupyter-chat::widget::${options.model.name}`;
+
+    // Track mouse position to detect text selection vs click
+    let mouseDownPos: { x: number; y: number } | null = null;
+
+    this.node.addEventListener('mousedown', (event: MouseEvent) => {
+      mouseDownPos = { x: event.clientX, y: event.clientY };
+    });
+
     this.node.addEventListener('click', (event: MouseEvent) => {
       const target = event.target as HTMLElement;
+
+      // Don't focus if something already has focus (e.g., input is focused)
       if (this.node.contains(document.activeElement)) {
         return;
       }
 
       const message = target.closest(`.${MESSAGE_CLASS}`);
       if (message) {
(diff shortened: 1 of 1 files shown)


brichet:
Thanks @veeceey for your interest, and @nakul-py for the quick review.
I agree that this issue is already fixed with #268, and I can't reproduce it with jupyter-ai:

[record-2026-02-09_09.13.09.webm](https://github.com/user-attachments/assets/b78b83c8-0b9a-40f8-a2d8-5de79ff74d8a)

brichet:
I'm closing this PR as already fulfilled by #268.

PR #363, opened 2026-02-18
Add `IMessagePreambleRegistry` extension point and `IMessageMetadata` interface

andrii-i:
### Description

Adds a message preamble registry (`IMessagePreambleRegistry`) and `IToolCall` type support to `jupyter-chat`.

The preamble registry is an extension point that allows JupyterLab extensions to register React components rendered above the message body, after the header. It follows the same pattern as the existing footer registry (Token + interface + implementation + plugin), including the same `IMessageContent`-based props signature. Components render vertically in insertion order. This enables downstream packages to plug UI into chat messages above the message body — for example, [jupyter-ai-acp-client#12](https://github.com/jupyter-ai-contrib/jupyter-ai-acp-client/pull/12) uses it to display real-time tool call status for AI agent actions.

Adds empty `IMessageMetadata` interface to be augmented by extensions and `metadata?: IMessageMetadata` field on `IMessageContent` (TypeScript) and `Message` (Python), providing first-class support for arbitrary metadata on chat messages.

### Tool Call UI (implementation) Preview
Preview from https://github.com/jupyter-ai-contrib/jupyter-ai-acp-client/pull/12 that uses the message preamble registry:
* Simple read
read_justfile
* Multiple agents
multiple_agents

### Tests

Unit tests for the preamble registry covering: empty initialization, component addition, insertion order preservation, and defensive copy (mutating the returned array does not affect internal state).

### Reviewer instructions

- Run `jlpm build`, ensure compiles clean
- Run `jlpm test`, ensure preamble registry tests pass
- Verify that `MessagePreambleComponent` renders between header and body in `messages.tsx`. To test end-to-end, install [jupyter-ai-acp-client#12](https://github.com/jupyter-ai-contrib/jupyter-ai-acp-client/pull/12) alongside this branch. [jupyter-ai-devrepo](https://github.com/jupyter-ai-contrib/jupyter-ai-devrepo) can simplify the multi-repo setup.

### DIFF
--- a/packages/jupyter-chat/src/message.ts
+++ b/packages/jupyter-chat/src/message.ts
@@ -9,7 +9,7 @@ import {
   IAttachment,
   IMessageContent,
   IMessage,
-  IToolCall,
+  IMessageMetadata,
   IUser
 } from './types';
 
@@ -71,8 +71,8 @@ export class Message implements IMessage {
   get stacked(): boolean | undefined {
     return this._content.stacked;
   }
-  get tool_calls(): IToolCall[] | undefined {
-    return this._content.tool_calls;
+  get metadata(): IMessageMetadata | undefined {
+    return this._content.metadata;
   }
 
   /**
--- a/packages/jupyter-chat/src/types.ts
+++ b/packages/jupyter-chat/src/types.ts
@@ -62,15 +62,18 @@ export interface IConfig {
 }
 
 /**
- * A tool call associated with an AI agent message.
+ * An empty interface to describe optional metadata attached to a chat message.
+ * Extensions can augment this interface to add custom fields:
+ *
+ * ```ts
+ * declare module '@jupyter/chat' {
+ *   interface IMessageMetadata {
+ *     myField?: MyType;
+ *   }
+ * }
+ * ```
  */
-export interface IToolCall {
-  tool_call_id: string;
-  title: string;
-  kind?: string;
-  status?: string;
-  raw_output?: unknown;
-}
+export interface IMessageMetadata {} /* eslint-disable-line @typescript-eslint/no-empty-object-type */
 
 /**
  * The chat message description.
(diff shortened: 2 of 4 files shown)


andrii-i:
Hi @brichet, other maintainers. Could you please give this a look?

dlqqq (approved):
These changes look good to me. The message preamble makes a lot of sense since we already have a message footer. My only minor nitpick is that the naming convention is a bit awkward, but it makes sense given that the username & avatar is already the "header", so we should call this component the "preamble".

If I understand correctly, I think the main area of discussion concerns the addition of the `tool_calls` field to the message model. IMO, there are a few good reasons for us to add this:

1. Jupyter Chat's de facto purpose is to serve as a chat library for our AI extensions, i.e. Jupyter AI & JupyterLite AI. We should feel free to add general AI-relevant fields to the message model to make it easier for our extensions to append metadata to AI response messages. It may go unused by consumers, which is fine because empty keys occupy minimal memory/disk relative to the message contents.

2. The tool call object schema is well standardized now so we're not locking consumers into any specific implementation by adding the `tool_calls` field here. The same schema is being used in LangChain and other AI libraries/providers.

3. While we could add some extension point for making the message model configurable, this introduces several new design questions to work through. What happens if two different extensions need to each add fields to the message model? How can we still preserve type safety without just making the message an arbitrary dictionary? What happens if a user uninstalls an extension that adds a custom field, then opens an old chat file containing that custom field in its messages? IMO, the proposed approach of adding universally helpful _but optional_ fields makes more sense pragmatically. We can update this in the future before the v1.0 release if needed.

andrii-i:
Thanks @brichet and @SylvainCorlay. Updated the PR to follow the proposed generic metadata approach. Please give it another look now.

dlqqq:
TYSM!

PR #364, opened 2026-02-19
feat: add delete chat button

### LINKED ISSUE #328: Chat Delete button
@jtpio @brichet Should we need a separate Delete chat button for every chat like we have Rename, Close chat butoons and Where?

erkin98:
Closes #328

This adds a way to delete chats directly from the side panel (via a trash icon in the section toolbar) and from the file browser context menu.

When you click the trash icon, a confirmation dialog pops up before anything gets deleted. The actual deletion goes through `contents.delete()`, and the existing `fileChanged` handler already takes care of refreshing the chat list and cleaning up the side panel section — so no extra cleanup logic was needed.

I followed the same patterns already used by the rename feature: optional callback passed through `MultiChatPanel` → `ChatSection`, guarded with an `if (options.deleteChat)` check, and using `Dialog.warnButton` for the confirmation (standard JupyterLab approach for destructive actions).

**What changed:**
- `token.ts` — added `deleteChat` command ID
- `multichat-panel.tsx` — delete button with confirmation dialog, disposal on success
- `index.ts` — command registration + callback wiring
- `commands.json` — context menu entry for `.chat` files

**Testing:**
- Open the side panel with a few chats, click the trash icon, confirm → chat gets removed
- Cancel the dialog → nothing happens
- Right-click a `.chat` file in the file browser → "Delete chat" option works
- Delete a chat that's also open in the main area → should clean up properly

### DIFF
--- a/packages/jupyter-chat/src/widgets/multichat-panel.tsx
+++ b/packages/jupyter-chat/src/widgets/multichat-panel.tsx
@@ -8,10 +8,11 @@
  * Originally adapted from jupyterlab-chat's ChatPanel
  */
 
-import { InputDialog } from '@jupyterlab/apputils';
+import { Dialog, InputDialog, showDialog } from '@jupyterlab/apputils';
 import {
   addIcon,
   closeIcon,
+  deleteIcon,
   HTMLSelect,
   launchIcon,
   PanelWithToolbar,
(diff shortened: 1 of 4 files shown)


brichet:
Thanks @erkin98 for working on this, and sorry for the late review.

IMO, this `delete` button makes sense for file based chats, and should be defined in `jupyterlab-chat`, not in `@jupyter/chat`.
Hopefully #445 will help adding this button to the toolbar from `jupyterlab-chat`.

PR #378, opened 2026-02-27
[FIX] Message toolbar buttons overlaps with text.

### LINKED ISSUE #375: Message action toolbar overlaps message text on hover
### Problem

When hovering over a message in Jupyter Chat, the message action toolbar can appear overlaid on top of the message text. This reduces text readability, makes the action buttons hard to notice, and gives the UI a broken appearance.

Image

### Proposed Solution

Redesign the message action toolbar so it doesn't interfere with text readability. Two possible approaches:

1. Position the toolbar so it never overlaps message text (e.g., above the message bubble or in a dedicated gutter)
2. If overlapping is unavoidable, give the toolbar a solid/opaque background so buttons are clearly distinguishable from the text beneath (see screenshot below as reference)

Image

nakul-py:
Adding solid background to message toolbar buttons and change their position to bottom right corner of the message.

https://github.com/user-attachments/assets/30d43bde-ab41-4b32-a10f-b724b29e7931

- Closes #375


brichet:
Don't know if we want a grey background on all messages.

Screenshot from 2026-03-09 13-32-51

@andrii-i any opinion on it ?

andrii-i (commented):
@nakul-py thank you for working on this. 

@nakul-py @brichet I think it's a clear improvement vs current state where icon has no background and still overlaps with text. As I mention in #375, ideally we'd position the toolbar so it never overlaps with message text (e.g., above the message bubble or in a dedicated gutter) but If overlapping is unavoidable, giving the toolbar a solid/opaque background so buttons are clearly distinguishable from the text beneath would also be an improvement.

Without this PR: 
Image

With this PR:
Image

brichet (approved):
Thanks for your patience @nakul-py, it looks good to me.

@andrii-i what do you think ?

brichet:
@andrii-i I merge this PR; it was approved quite a while ago. We can always update it later if necessary.

PR #396, opened 2026-04-03
Drop lerna dependency

brichet:
This PR drop the lerna dependency.

For the reference, it is already done or WIP in main Jupyter projects:
- https://github.com/jupyterlite/jupyterlite/pull/1843
- https://github.com/jupyterlab/jupyterlab/pull/18444
- https://github.com/jupyter/notebook/pull/7882


scripts/bump_version.py
```diff
@@ -54,10 +52,9 @@ def increment_version(current, spec):
 
 
 @click.command()
-@click.option("--force", default=False, is_flag=True)
 @click.option("--skip-if-dirty", default=False, is_flag=True)
 @click.argument("spec", nargs=1)
-def bump(force, skip_if_dirty, spec):
+def bump(skip_if_dirty, spec):
```
jtpio, line 57: Wondering if we could put this script under tests on CI, just checking if we are able to bump versions?

This could be done separately though, if we are confident the changes still ensure the current release process works as expected.

jtpio (approved):
Thanks!

Looks like it picked a small conflict from the latest patch release.

Otherwise the diff looks good.

scripts/bump_version.py
brichet: Isn't it already the case with `check_release` workflow ?
https://github.com/jupyterlab/jupyter-chat/actions/runs/23946018613/job/69842286553#step:4:300

jtpio: Yeah by default it tests bumping as `next` iirc. I meant checking that passing another value like the version number also works fine.

brichet: I opened #399

PR #397, opened 2026-04-03
Rerender the message when mime model is updated

brichet:
This PR fix the re-rendering of the message when the mime model is updated.


jtpio (approved):
Thanks!

PR #405, opened 2026-04-09
Fix chat input when it's an attachment

Yahiewi:
I came across this issue while working on a [jupyterlite-ai PR](https://github.com/jupyterlite/ai/pull/306).
The issue is that when the user message contains only an attachment and no text, it can't be sent using the 'Enter' key but can be sent with the 'Send Message' button.
Looking closely at the code the `inputExists` constant in the `ChatInput` component only looked at the text input and ignored the attachments unlike the [`SendButton`](https://github.com/jupyterlab/jupyter-chat/blob/main/packages/jupyter-chat/src/components/input/buttons/send-button.tsx#L35) component which handles this correctly.

### DIFF
--- a/packages/jupyter-chat/src/components/input/chat-input.tsx
+++ b/packages/jupyter-chat/src/components/input/chat-input.tsx
@@ -107,7 +107,7 @@ export function ChatInput(props: ChatInput.IProps): JSX.Element {
     };
   }, [inputToolbarRegistry]);
 
-  const inputExists = !!input.trim() || attachments.length > 0;
+  const inputExists = !!input.trim();
 
   /**
    * `handleKeyDown()`: callback invoked when the user presses any key in the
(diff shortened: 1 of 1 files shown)


brichet (commented):
Thanks @Yahiewi for looking into it.

It allows adding new lines in an empty message, that was not allowed before. Maybe we should keep the same behavior, whether there is an attachment or not.

https://github.com/user-attachments/assets/dbf26511-dfa7-4b15-9c9d-d29623120108

Yahiewi:
Thanks @brichet for pointing this out.
Here are the changes I made to fix this:
-I reverted inputExists to `const inputExists = !!input.trim();` 
-Defined a new ```
const isSendCombination =
      (sendWithShiftEnter && event.shiftKey) ||
      (!sendWithShiftEnter && !event.shiftKey);```
-Changed the condition for blocking the event to `if (!inputExists && (!isSendCombination || attachments.length === 0))` which basically means: if there is no text, block the key press unless it's a send command and there is an attachment.

brichet (approved):
Thanks @Yahiewi, LGTM

PR #409, opened 2026-04-21
Add a placeholder widget in side panel

### LINKED ISSUE #408: Placeholder in empty side panel
### Problem

When no chat is opened in the side panel, the panel is empty except for the top toolbar, with a button to create a chat and a select to open an existing one.

We could add a placeholder in this area, like in the TOC or notebook properties panels.

 

### Proposed Solution

- add a default placeholder (customizable) in the side panel when no chat is opened (`@jupyter/chat`)
- add an extension point in `jupyterlab-chat` to allow third party extensions to customize it.

### Additional context

Related discussion in Zulip https://jupyter.zulipchat.com/#narrow/channel/475130-jupyter-ai/topic/jupyter-chat/near/587801832

Related issue:
- https://github.com/jupyterlab/jupyter-ai/issues/1556

brichet:
Add a placeholder to the empty chat panel.

image

It add a "+" button and the list of available chats. This panel is displayed by default, and when the current chat is closed.

There is also an extension point, the plugin in `jupyterlab-chat` has an optional dependency on `IChatPlaceholderFactory` that can be provided by an extension, and propagated to the chat side panel.

Fixes #408

### DIFF
--- a/packages/jupyter-chat/src/tokens.ts
+++ b/packages/jupyter-chat/src/tokens.ts
@@ -5,8 +5,9 @@
 
 import { IWidgetTracker, MainAreaWidget } from '@jupyterlab/apputils';
 import { Token } from '@lumino/coreutils';
+import { Widget } from '@lumino/widgets';
 
-import { ChatWidget } from './widgets';
+import { ChatWidget, Placeholder } from './widgets';
 import { IChatModel } from './model';
 
 /**
@@ -27,4 +28,27 @@ export type IChatTracker = IWidgetTracker<ChatWidget | MainAreaChat>;
 export const IChatTracker = new Token<IChatTracker>(
   '@jupyter/chat:IChatTracker',
   'The chat widget tracker'
+);
+
+/**
+ * The interface for the placeholder factory.
+ */
+export interface IChatPlaceholderFactory {
+  /**
+   * Create a placeholder widget for the multi-chat panel.
+   *
+   * @param props - the props passed to the placeholder.
+   * @returns a widget to display as placeholder.
+   */
+  create(props: Placeholder.IProps): Widget;
+}
+
+/**
+ * The token for the placeholder factory.
+ * Not provided by default — extensions can provide it to replace the default
+ * placeholder shown when no chat is open in the multi-chat panel.
+ */
+export const IChatPlaceholderFactory = new Token<IChatPlaceholderFactory>(
+  '@jupyter/chat:IChatPlaceholderFactory',
+  'The placeholder factory for the multi-chat panel'
 );
--- a/packages/jupyter-chat/src/widgets/placeholder.tsx
+++ b/packages/jupyter-chat/src/widgets/placeholder.tsx
@@ -3,6 +3,7 @@
  * Distributed under the terms of the Modified BSD License.
  */
 
+import { IRenderMime } from '@jupyterlab/rendermime';
 import {
   addIcon,
   ReactWidget,
@@ -55,6 +56,10 @@ export namespace Placeholder {
      * An optional signal emitting when the chat list changes.
      */
     chatNamesChanged?: ISignal<any, { [name: string]: string }>;
+    /**
+     * An optional translation bundle.
+     */
+    trans?: IRenderMime.TranslationBundle;
   }
 }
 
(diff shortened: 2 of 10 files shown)


brichet:
I wonder if we want an extension point for it, actually.

brichet:
> I wonder if we want an extension point for it, actually.

I added a token for a placeholder factory, that can be provided by an extension.

krassowski:
An alternative take:

image

For reference, VSCode

image


PyCharm - first start

image


PyCharm - once configured

image


Chat GPT

image

Claude Web

image

krassowski:
Do we have the timestamps information to display?

packages/jupyter-chat/src/tokens.ts
```diff
@@ -28,3 +29,26 @@ export const IChatTracker = new Token<IChatTracker>(
...
+ * The interface for the placeholder factory.
+ */
+export interface IChatPlaceholderFactory {
+  /**
+   * Create a placeholder widget for the multi-chat panel.
+   *
+   * @param props - the props passed to the placeholder.
+   * @returns a widget to display as placeholder.
+   */
+  create(props: Placeholder.IProps): Widget;
```
krassowski, line 43: Is the `Placeholder.IProps` easily importable downstream?

One option to ensure that would be


```suggestion
  create(props: IChatPlaceholderFactory.ICreateProps): Widget;
}

export namespace IChatPlaceholderFactory {
  export interface ICreateProps extends Placeholder.IProps {}
}
```

another would be exporting `Placeholder` from the main `index.ts`.

Or maybe nothing is needed here, I am not sure.

packages/jupyter-chat/src/widgets/placeholder.tsx
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * The placeholder namespace.
+ */
+export namespace Placeholder {
+  /**
+   * The options of the placeholder widget.
+   */
+  export interface IProps {
```
krassowski, line 40: Should it also include translator as optional and make use of it?

FYI the latest `@jupyter/eslint-plugin` can highlight strings where `trans.__()` is missing (https://eslint-plugin.readthedocs.io/en/latest/rules/no-untranslated-string/)

packages/jupyter-chat/src/widgets/placeholder.tsx
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * The initial chat names.
+     */
+    chatNames: { [name: string]: string };
```
krassowski, line 44: Maybe including an object for each chat so that we can expand it easily later (with information like creation date for now, in future I guess other metadata might be useful for downstreams)?

packages/jupyter-chat/src/widgets/placeholder.tsx
brichet: > FYI the latest @jupyter/eslint-plugin can highlight strings where trans.__() is missing (https://eslint-plugin.readthedocs.io/en/latest/rules/no-untranslated-string/)

Thanks for the heads up, we should include it in this project.

packages/jupyter-chat/src/tokens.ts
brichet: > Is the Placeholder.IProps easily importable downstream?

Yes

```ts
import { Placeholder } from '@jupyter/chat';
const props: Placeholder.IProps = {...}
```

Not sure that we need to export `IChatPlaceholderFactory`.

packages/jupyter-chat/src/widgets/placeholder.tsx
brichet: The placeholder widget has now an optional translation bundle.

packages/jupyter-chat/src/widgets/placeholder.tsx
krassowski: I see that `getChatNames` is a pattern used thought public API so I guess the metadata could be passed as a new property mapping chat to extra info. Resolving for now.

krassowski (approved):
LGTM!

brichet:
Thanks @krassowski for the review.

packages/jupyter-chat/src/widgets/placeholder.tsx
brichet: Yes, it could be nice to be able to extend the properties, for example to add a title to the chats https://github.com/jupyterlite/ai/pull/327, instead of only the name.

PR #410, opened 2026-04-24
Bump eslint to v9 and use @jupyter/eslint-plugin

brichet:
Use `@jupyter/eslint-plugin` https://github.com/jupyterlab/eslint-plugin

Similar to https://github.com/jupyterlab/extension-template/pull/141/changes


brichet:
Bumping the CI

.github/workflows/build.yml
```diff
@@ -25,19 +25,17 @@ jobs:
       - name: Install dependencies
         run: python -m pip install -U "jupyterlab>=4.0.0,<5"
 
-      - name: Lint the packages
-        run: |
+      - name: Build the core package
+        run:  |
```
krassowski:
```suggestion
        run: |
```

krassowski:
I think it errors because the `build:core` only builds one package rather than all packages/

brichet:
> I think it errors because the `build:core` only builds one package rather than all packages/

Thanks, maybe we should run the lint after `build_extension`.

package.json
```diff
@@ -46,19 +46,21 @@
     "watch": "jlpm workspace jupyterlab-chat-extension run watch"
   },
   "devDependencies": {
+    "@eslint/js": "^9.0.0",
+    "@jupyter/eslint-plugin": "^0.0.3",
```
krassowski:
```suggestion
    "@jupyter/eslint-plugin": "^0.0.5",
```

If you wish to have the translation and token format checks, I would recommend `v0.0.5`

krassowski (approved):
🚀

brichet:
Thanks @krassowski for the review.

PR #413, opened 2026-04-27
Create chat ID without relying on dirty state

brichet:
This PR avoid using the `dirty` state of the shared document (chat in `jupyterlab-chat`) to create the chat ID.

This is related to future changes in [`jupyterlab`](https://github.com/jupyterlab/jupyterlab/pull/18739), that won't set the `dirty` state anymore when initializing.


brichet:
It seems that jupyter-collaboration still set the `dirty` flag to `false` on [first load](https://github.com/jupyterlab/jupyter-collaboration/blob/092b288555a0e9c704c390993e444bda19a9d188/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L174), so I wonder if this PR is still required.

brichet:
This PR is not relevant anymore, the `dirty` state is still set to `true` during initialization and `false` after initialization (see https://github.com/jupyterlab/jupyterlab/pull/18739#discussion_r3298811216). We can still assume that the the document is loaded when the `dirty` is set to `false` the first time.

On the other hand, there is #439 that can be useful to handle chat readiness from the frontend only (e.g. webRTC collaboration).

PR #415, opened 2026-04-28
Fix chat sidepanel toolbar overflow behavior

### LINKED ISSUE #330: UI bug in the chat toolbar
https://github.com/user-attachments/assets/e8d511f7-0d3f-4e9c-8c0f-87f18466a839

As you can see here when the sidepanel which contains the toolbar is shrinked, the UI of the toolbar can break with some buttons that fall out of the container. I think the easiest fix for this would be to make the toolbar horizontally scrollable.

MUFFANUJ:
Ref/ closes https://github.com/jupyterlite/ai/issues/330

This updates the chat sidepanel toolbar to use responsive overflow behavior that keeps visible toolbar actions in place by default and only moves actions that exceed the available width into the ellipsis menu. It aligns the layout with JupyterLab toolbar patterns using the reactive toolbar API.


packages/jupyter-chat/style/chat.css
```diff
@@ -148,8 +148,13 @@
...
   overflow: hidden;
   text-overflow: ellipsis;
+  white-space: nowrap;
   text-transform: uppercase;
   color: var(--jp-ui-font-color1);
   font-size: var(--jp-ui-font-size0);
   font-weight: bold;
 }
+
+.jp-chat-sidepanel-widget-toolbar > .jp-Toolbar-responsive-opener {
```
brichet:
```suggestion
.jp-chat-sidepanel-widget-toolbar > .jp-Toolbar-item.jp-ToolbarButton:first-of-type {
```

In addition to a max-width to the title, to keep the buttons on the right.

packages/jupyter-chat/src/widgets/multichat-panel.tsx
```diff
@@ -508,14 +525,15 @@ export namespace MultiChatPanel {
...
-class SidePanelWidget extends PanelWithToolbar {
+class SidePanelWidget extends ReactivePanelWithToolbar {
   constructor(options: SidePanelWidget.IOptions) {
     super();
     this._chatWidget = options.widget;
     this._displayName = options.displayName ?? options.widget.model.name;
     const trans = options.trans;
 
     this.addClass(SIDEPANEL_WIDGET_CLASS);
+    this.toolbar.addClass('jp-SidePanel-toolbar');
```
brichet, line 543: Is this necessary ?

packages/jupyter-chat/src/widgets/multichat-panel.tsx
```diff
@@ -45,6 +46,22 @@ const OPEN_SELECT_CLASS = 'jp-chat-open';
 const SIDEPANEL_WIDGET_CLASS = 'jp-chat-sidepanel-widget';
 const TOOLBAR_CLASS = 'jp-chat-sidepanel-widget-toolbar';
 
+/**
+ * A panel widget with a reactive toolbar.
```
brichet, line 53: We should probably add a comment to remove this class when bumping to `@jupyterlab/ui-components 4.6.0`, after https://github.com/jupyterlab/jupyterlab/pull/18824

brichet (commented):

On my side the ellipsis button is displayed too early, and the title disappears also very early (it should not disappear at all IMO).

[record-2026-05-03_21.53.15.webm](https://github.com/user-attachments/assets/ae206b89-2b34-4db1-9354-ba27871c1bac)

I wonder if we should not keep a max width to the title, and add a margin-left auto to the first button (instead of the ellipsis button).

packages/jupyter-chat/src/widgets/multichat-panel.tsx
MUFFANUJ: Yes, for now this is necessary. `SidePanelWidget` extends `Panel` (not `SidePanel`), so its toolbar does not automatically get side-panel toolbar styles. Adding `jp-SidePanel-toolbar` applies the standard sidebar toolbar sizing/behavior from `@jupyterlab/ui-components` (notably the side-panel toolbar height), which keeps the layout consistent and avoids early overflow behavior. I added a TODO to drop this workaround once we can rely on `@jupyterlab/ui-components >= 4.6.0`.

brichet:
@MUFFANUJ on my side the ellipsis button seems to be revealed too early.

Maybe another way to fix it could be to add a [spacer](https://github.com/jupyterlab/jupyterlab/blob/3554aafaf2d8b2bca167fb34a28a994799f05bd1/packages/ui-components/src/components/toolbar.tsx#L775) between the title and the buttons, and remove the `flex: 1 1 0%;` from CSS rules of the title. I have something that "kind" of work if the title has a maximum size too (there are still some glitches to fix, though).

[record-2026-06-17_19.10.13.webm](https://github.com/user-attachments/assets/ff26b3a1-13a0-4f76-9781-4111f8de38f9)

brichet:
@MUFFANUJ do you think you'll be able to finish this PR soon?
It would be nice to include it in the next release.

MUFFANUJ:
Sure i will work on this later in the day i was occupied with other tasks and missed this one thanks for the ping!

packages/jupyter-chat/style/chat.css
```diff
@@ -147,10 +147,12 @@
 .jp-chat-sidepanel-widget-toolbar > .jp-chat-sidepanel-widget-title {
   display: block;
   align-content: center;
-  flex: 1 1 0%;
+  flex: 0 1 auto;
```
brichet: These flex properties are the default ones, do you think that we need to keep it ?

brichet (approved):
Thanks @MUFFANUJ, LGTM 👍 !

I just have a non blocking comment.

packages/jupyter-chat/style/chat.css
MUFFANUJ: We don't need it.

brichet:
Thanks @MUFFANUJ

PR #416, opened 2026-04-29
Restore send with selection

### LINKED ISSUE #303: Follow up UX improvments

brichet:
This PR restores the 'send with selection' feature in `@jupyter/chat`.
It allows to include text selection (from a notebook or editor widget) or active cell content.

There is also a setting in `jupyterlab-chat` to hide it if not expected.

- fixes #303 
- related to https://github.com/jupyterlab/jupyter-ai/issues/1536

### Discussion

Should we include the selection as an attachment instead ?
Currently is is added as `code` in the message.

[output.webm](https://github.com/user-attachments/assets/7fc84763-65b3-435f-a262-76ce496854e3)

### DIFF
--- a/packages/jupyterlab-chat-extension/schema/factory.json
+++ b/packages/jupyterlab-chat-extension/schema/factory.json
@@ -46,7 +46,7 @@
     "sendWithSelection": {
       "description": "Show the 'send with selection' button in the chat input.",
       "type": "boolean",
-      "default": true,
+      "default": false,
       "readOnly": false
     },
     "defaultDirectory": {
--- a/packages/jupyterlab-chat-extension/schema/factory.json
+++ b/packages/jupyterlab-chat-extension/schema/factory.json
@@ -43,6 +43,12 @@
       "default": false,
       "readOnly": false
     },
+    "sendWithSelection": {
+      "description": "Show the 'send with selection' button in the chat input.",
+      "type": "boolean",
+      "default": true,
+      "readOnly": false
+    },
     "defaultDirectory": {
       "description": "Default directory where to create and look for chat, the jupyter root directory if empty.",
       "type": "string",
--- a/docs/source/users/index.md
+++ b/docs/source/users/index.md
@@ -119,15 +119,30 @@ These settings includes:
 
 - **stackMessages**
 
-  Whether to stack consecutive messages from same user.\
+  Whether or not to stack consecutive messages from same user.\
   Default: true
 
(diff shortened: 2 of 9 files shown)


jtpio:
> Should we include the selection as an attachment instead ?

What was the previous behavior? Maybe we can align with the previous behavior as a first step, and then refine later based on user feedback?

packages/jupyterlab-chat-extension/schema/factory.json
```diff
@@ -43,6 +43,12 @@
       "default": false,
       "readOnly": false
     },
+    "sendWithSelection": {
+      "description": "Show the 'send with selection' button in the chat input.",
+      "type": "boolean",
+      "default": true,
```
jtpio: Wondering about the default value here.

Maybe it should be `false` by default, and have documentation to explain how to enable it?

Or automatically show it only if there is an active selection?

jtpio (approved):
Thanks!

Changes look good. Just wondering about the default for the button (left an inline comment).

packages/jupyterlab-chat-extension/schema/factory.json
brichet: > Maybe it should be false by default, and have documentation to explain how to enable it?

I don't have a strong opinion on it, looks also good to have it to `false` by default.

> Or automatically show it only if there is an active selection?

Not sure about that, that would ad some glitch I think.

brichet:
Thanks @jtpio, for the review, it's now set to `false`, and added to the documentation.

jtpio (approved):
Thanks!

PR #417, opened 2026-04-29
add error notification on failed cell drag drop

### LINKED ISSUE #254: Show error notification when cell drag fails
When dragging notebook cells fails (due to missing notebook tracker, cell not found, or other issues), show a user-visible error notification instead of only logging error to console (which a lot of users might not see / check).

> Since this entire method depends on `notebookTracker`, can this method just return `null` early if that is not available? The parent method should log a warning if this method ever returns `null`. May also be worth opening a `Notification.error()` popup for visibility.

_Originally posted by @dlqqq in https://github.com/jupyterlab/jupyter-chat/pull/248#discussion_r2183181570_

MUFFANUJ:
Fixes #254 

This PR adds `Notification.error(...)` messages in the chat widget when notebook cell drag-and-drop fails, instead of only logging to the console.


brichet:
Thanks @MUFFANUJ.
Is there an easy way to reproduce this kind of error ?

MUFFANUJ:
Yes, we could,
Go to  “Advanced Plugin Manager”.
Search for `jupyterlab-chat-extension:activeCellManager` and disable it ( and reload).
Open a notebook and open chat and Drag a notebook cell into chat input.

Expected result: you should get the error toast
`Failed to attach notebook cells: source notebook could not be located`.


https://github.com/user-attachments/assets/963c0208-623b-46aa-9778-a162f96adaaa

MUFFANUJ:
If we want to show it directly and not after clicking on the bell, we can add autoclose with false for error.

brichet (approved):
Thanks @MUFFANUJ, LGTM!

PR #419, opened 2026-04-29
Unify chat message submit logic across input handlers

### LINKED ISSUE #230: Sending a message logic is split across 3 areas
### Problem

There are 3 areas of the code where a message is being sent:

1. `sendWithSelection()` in `SendButton`
2. The `onClick()` handler in `SendButton`
3. The `handleKeyDown()` handler in `ChatInput`

Ideally, there should be a unified way to send a message that doesn't require changing all 3 of these areas.

### Proposed Solution

Explore refactoring the structure of the chat extension to pass references more easily.

MUFFANUJ:
Fixes #230 

This PR consolidates chat message submission into a single shared helper used by the send button, keyboard submit in ChatInput, and save-edit action. It keeps existing behavior unchanged while removing duplicated logic.


packages/jupyter-chat/src/components/input/submit-message.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Submit the current input value as a chat message.
+ */
+export async function submitInputMessage(
+  options: submitInputMessage.IOptions
+): Promise<void> {
+  const {
+    model,
+    chatCommandRegistry,
+    clearInputBeforeSend = false,
+    focusInputAfterSend = false
```
brichet: 1. I think that we can simplify this part, and remove `clearInputBeforeSend` and ` focusInputAfterSend`:    
 - never clear the focus before sending the message.
   IIRC it was a workaround when the send message was awaited, that has been fixed in https://github.com/jupyterlite/ai/issues/260, but I don't think that clearing the input was necessary
 - always focus on the input after sending the message, IMO this is what is expected

2. In addition, we could add an optional `body` argument, that would be used if provided, or default to `model.value`. That way, we could also use this submit function in https://github.com/jupyterlab/jupyter-chat/blob/24dae6826749fa29e45c61338a700073355bcfe4/packages/jupyter-chat/src/components/input/buttons/send-button.tsx#L124

brichet (commented):

brichet:
@MUFFANUJ do you think that we could also use this function with `sendWithSelection` too ?
https://github.com/jupyterlab/jupyter-chat/blob/24dae6826749fa29e45c61338a700073355bcfe4/packages/jupyter-chat/src/components/input/buttons/send-button.tsx#L124

Something like
```ts
await submitInputMessage({ model, chatCommandRegistry, body });
```

MUFFANUJ:
@brichet Thanks for pointing it out i missed this earlier i have added the suggestion in the latest commit.

MUFFANUJ:
Test failure does not seem related.

brichet (approved):
Thanks @MUFFANUJ, it looks good to me.

brichet:
@MUFFANUJ I merged main in your branch to include #444, that should fix the tests on placeholder.

PR #420, opened 2026-04-30
[Feature] Set file browser path as chat cwd metadata

Zsailer:
I like to tuck my chats away in a single folder (like .jupyter/chats) so they don't clutter my file tree. Problem is, when a server-side extension wants to know where I was actually working when I opened the chat, the only directory info it can find is where the chat file lives.

This writes the file browser's current path as a `cwd` key in the YChat metadata when a chat model is created. The write happens after the Y.js document finishes its initial sync so the server's state doesn't overwrite it.

From the server side, any extension with access to the YChat can read `ychat.get_metadata().get("cwd")` to find out where the user was browsing. Handy for AI agents, code assistants, or anything that benefits from knowing the user's working context.

### DIFF
--- a/packages/jupyterlab-chat-extension/src/index.ts
+++ b/packages/jupyterlab-chat-extension/src/index.ts
@@ -104,7 +104,8 @@ async function createChatModel(
   app: JupyterFrontEnd,
   contentProvider: ICollaborativeContentProvider,
   path?: string,
-  defaultDirectory?: string
+  defaultDirectory?: string,
+  filebrowser?: IDefaultFileBrowser | null
 ): Promise<MultiChatPanel.IOpenChatArgs> {
   const modelFactory = app.docRegistry.getModelFactory(
     'Chat'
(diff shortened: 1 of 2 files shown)


dlqqq:
Is this the right solution to the problem, or should we just revert Jupyter AI back to using the server's root directory / preferred directory as the CWD?

I don't see much benefit to scoping Jupyter AI's agents to a specific directory. I merged the PR that introduced the change since that was our original plan, but I think we should reconsider.

brichet:
I wonder if this metadata should be handled in [`jupyter-ai-router`](https://github.com/jupyter-ai-contrib/jupyter-ai-router/blob/6b78fe788ff5419112582d997852ca5f21e3e313/jupyter_ai_router/extension.py#L79) instead, as it seems to be really AI related information. 
But maybe I miss some other use cases where the initial working directory may be interesting to keep in the chat file.

In any case, we should probably provide a way to display and update this path. Relying solely on an initial state for persistent data could lead to errors.

Zsailer:
Closing this, as I think I need to re-think the approach here quite a bit. Thanks @brichet !

PR #421, opened 2026-04-30
[Bug] Launcher ignores configured defaultDirectory when creating chats

Zsailer:
When a `defaultDirectory` is configured (via settings or server config), chats created from the side panel correctly land in that directory. But chats created from the launcher, menu, or command palette ignore it and use the file browser's current directory instead.

The `createChat` command was only checking `inSidePanel` to decide whether to use the configured directory. Now it also checks whether a `defaultDirectory` is configured -- if one exists, it's used regardless of where the chat was created from. The file browser fallback only kicks in when no default is configured.

Includes a UI test that configures a default directory, creates a chat from the launcher, and verifies it lands in the right place.


brichet:
Thanks @Zsailer for working on this.

This PR seems to revert the behavior introduced in #238, so I believe that it was expected.

cc. @andrii-i @ellisonbg

Zsailer:
Closing, since I don't think this was the right approach.

Issue #411, opened 2026-04-24
'Permission denied: untitled.chat' - `c.ServerApp.root_dir = "/"`


jhgoebbert:
## Description

Creating a new chat fails because `untitled.chat` is written to `root_dir` which might not be writeable.

## Reproduce

1. Start the jupyterlab with a config.py including `c.ServerApp.root_dir = "/"`
2. Click on 'Create a new chat'
3. Give the new chat a name and click on Ok.
4. Nothing will happen - but in stderr you can read `wrote error:
    - ```
       Permission denied: untitled.chat'`
       PermissionError: [Errno 13] Permission denied: '/untitled.chat'
       raise HTTPError(403, "Permission denied: %s" % path) from e
       tornado.web.HTTPError: HTTP 403: Forbidden (Permission denied: untitled.chat)
       ```
<details>
<summary>Unfold for complete error message</summary>

```
[W 2026-04-24 14:09:22.563 ServerApp] wrote error: 'Permission denied: untitled.chat'
    Traceback (most recent call last):
      File "/p/software/default/stages/2026/software/jupyter-server/2.17.0-GCCcore-14.3.0/lib/python3.13/site-packages/jupyter_server/services/contents/fileio.py", line 256, in perm_to_403
        yield
      File "/p/software/default/stages/2026/software/jupyter-server/2.17.0-GCCcore-14.3.0/lib/python3.13/site-packages/jupyter_server/services/contents/fileio.py", line 246, in atomic_writing
        with atomic_writing(os_path, *args, **kwargs) as f:
             ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/p/software/default/stages/2026/software/Python/3.13.5-GCCcore-14.3.0/lib/python3.13/contextlib.py", line 141, in __enter__
        return next(self.gen)
      File "/p/software/default/stages/2026/software/jupyter-server/2.17.0-GCCcore-14.3.0/lib/python3.13/site-packages/jupyter_server/services/contents/fileio.py", line 140, in atomic_writing
        fileobj = open(path, "wb", **kwargs)  # noqa: SIM115
    PermissionError: [Errno 13] Permission denied: '/untitled.chat'

    The above exception was the direct cause of the following exception:

    Traceback (most recent call last):
      File "/p/software/default/stages/2026/software/tornado/6.5.1-GCCcore-14.3.0/lib/python3.13/site-packages/tornado/web.py", line 1848, in _execute
[...]

PR #422, opened 2026-05-02
replaces `/` by `PreferredPath + .jupyter/` as the default directory if available

jhgoebbert:
This PR replaces `/` by `PreferredPath + .jupyter/` (if available) as the default directory for chat-files if defaultDirectory is not set or empty.  
It is a possible solution for the issue https://github.com/jupyterlab/jupyter-chat/issues/411

### DIFF
--- a/packages/jupyterlab-chat-extension/src/index.ts
+++ b/packages/jupyterlab-chat-extension/src/index.ts
@@ -43,10 +43,7 @@ import {
   showErrorMessage
 } from '@jupyterlab/apputils';
 import { IEditorLanguageRegistry } from '@jupyterlab/codemirror';
-import {
-  PathExt,
-  PageConfig
-} from '@jupyterlab/coreutils';
+import { PathExt, PageConfig } from '@jupyterlab/coreutils';
 import { DocumentRegistry } from '@jupyterlab/docregistry';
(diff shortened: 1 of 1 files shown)


packages/jupyterlab-chat-extension/src/index.ts
```diff
@@ -581,10 +581,19 @@ const chatCommands: JupyterFrontEndPlugin<void> = {
         // dir. Create new chat in file browser cwd if created from main
         // area (launcher, menu, palette).
         if (targetDirectory === undefined) {
+          const hasPreferredPath = PageConfig.getOption('preferredPath');
+          const preferredPath = hasPreferredPath
```
brichet, line 585: Should we check that `preferredPath` exists ?

packages/jupyterlab-chat-extension/src/index.ts
```diff
@@ -581,10 +581,19 @@ const chatCommands: JupyterFrontEndPlugin<void> = {
...
+            ? hasPreferredPath + '/.jupyter/'
```
brichet, line 586: `.jupyter` is more for configuration file I think, not sure if this is relevant for chat files.

brichet (commented):
Thanks @jhgoebbert for working on this.

Do you think that you can add a test on this change ?

PR #424, opened 2026-05-04
Fix flaky tests by avoiding duplicate file names

brichet:
Sometime the tests are not passing, because the file names are the same across tests running in parallel.

This PR fix it by using a different name for the chat file of each UI test.


brichet:
Merging it as it is only about file name used in tests.

PR #425, opened 2026-05-05
Add log when building packages

brichet:
Follow up #396, to add some logs when building the packages.


brichet:
Merging it as it does not touch the codebase and logs the building package https://github.com/jupyterlab/jupyter-chat/actions/runs/25362284741/job/74364742378?pr=425#step:5:51

PR #432, opened 2026-05-11
[Bug] Pass defaultDirectory when opening chats via openChat command

Zsailer:
The `openChat` command hardcodes `undefined` for the `defaultDirectory` parameter in `createChatModel`. This causes `getDisplayName` to show the full path (e.g. `.jupyter/chats/untitled`) instead of the short name (`untitled`) when a chat is restored on page refresh.

The `chatPanel`'s own `createModel` callback already passes the correct value from `widgetConfig` — this just makes the `openChat` command consistent.

I'm building a custom Lumino application that reuses the chat panel plugin and noticed the display name was wrong after page refresh. Traced it to this one-liner.

### DIFF
--- a/packages/jupyterlab-chat-extension/src/index.ts
+++ b/packages/jupyterlab-chat-extension/src/index.ts
@@ -825,7 +825,12 @@ const chatCommands: JupyterFrontEndPlugin<void> = {
                 return true;
               }
 
-              const openChatArgs = await createChatModel(app, drive, filepath);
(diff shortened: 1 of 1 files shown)


brichet:
Thanks @Zsailer for working on this.
On my side it seems that the issue is not solved with this PR.

I think that the config may not be filled with the settings value when using it.
Maybe we should:
- wait for the settings to be loaded before using the `widgetConfig` object
- and/or update the title when the config change (at least the default directory)

PR #433, opened 2026-05-11
 Add placeholder widget for empty chat files to provide customizable widgets.

bhavana-nair:
Introduce a factory-based extension point that allows extensions to render interactive widgets when a chat has no messages. The factory is consumed as an optional dependency in both the main-area chat widget and the side-panel chat.

UI customizable for any extension using this token. 
https://github.com/user-attachments/assets/0efec274-a179-4d69-8fee-950d4b41b431

### DIFF
--- a/packages/jupyter-chat/src/components/chat.tsx
+++ b/packages/jupyter-chat/src/components/chat.tsx
@@ -179,10 +179,10 @@ export namespace Chat {
      */
     welcomeMessage?: string;
     /**
-     * Optional factory to create message suggestions shown when the chat has
-     * no messages. 
+     * Optional factory to create chat body placeholder shown when the chat has
+     * no messages.
      */
-    messageSuggestionsFactory?: import('../tokens').IMessageSuggestionsFactory;
+    chatBodyPlaceholderFactory?: import('../tokens').IChatBodyPlaceholderFactory;
     /**
      * The area where the chat is displayed.
      */
--- a/packages/jupyter-chat/src/components/chat.tsx
+++ b/packages/jupyter-chat/src/components/chat.tsx
@@ -179,6 +179,11 @@ export namespace Chat {
      */
     welcomeMessage?: string;
     /**
+     * Optional factory to create message suggestions shown when the chat has
+     * no messages. 
+     */
+    messageSuggestionsFactory?: import('../tokens').IMessageSuggestionsFactory;
+    /**
      * The area where the chat is displayed.
      */
     area?: ChatArea;
--- a/packages/jupyter-chat/src/components/chat.tsx
+++ b/packages/jupyter-chat/src/components/chat.tsx
@@ -29,6 +29,7 @@ import {
   IMessagePreambleRegistry
 } from '../registers';
 import { ChatArea } from '../types';
+import { IChatBodyPlaceholderFactory } from '../tokens';
 
 export function ChatBody(props: Chat.IChatProps): JSX.Element {
   const { model } = props;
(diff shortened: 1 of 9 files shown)


packages/jupyterlab-chat-extension/src/index.ts
```diff
@@ -16,6 +16,7 @@ import {
   IMessageFooterRegistry,
   IMessagePreambleRegistry,
   IChatPlaceholderFactory,
+  IMessageSuggestionsFactory,
```
dlqqq: Let's consider renaming this to something that contains the word "placeholder". Consumers don't have to use this to show message suggestions for AI - consumers can show whatever they want here. For example, can we can name this `IChatBodyPlaceholderFactory`?

Also, if we're OK with it, I would suggest renaming both (technically a breaking change):

- chat panel placeholder : `IChatPanelPlaceholderFactory`
- chat body placeholder : `IChatBodyPlaceholderFactory`

dlqqq (commented):
Will complete full review later.

Wanted to add some context: This PR adds a placeholder that shows inside of an empty chat *file*. The existing placeholder only shows inside of an empty chat panel (with no chat file open).

dlqqq:
Also run `jlpm lint` a couple times & commit the new changes to fix CI.

packages/jupyterlab-chat-extension/src/index.ts
bhavana-nair: Renamed IMessageSuggestionsFactory placeholder to IChatBodyPlaceholderFactor.

dlqqq (approved):

dlqqq:
cc @brichet @jtpio Can you one of you offer a review on this PR? If it looks good, can we get a v0.22 release for this?

packages/jupyter-chat/src/components/chat.tsx
```diff
@@ -178,6 +178,11 @@ export namespace Chat {
      * The welcome message.
      */
     welcomeMessage?: string;
+    /**
+     * Optional factory to create chat body placeholder shown when the chat has
+     * no messages.
+     */
+    chatBodyPlaceholderFactory?: import('../tokens').IChatBodyPlaceholderFactory;
```
brichet: Is there a reason to import the type in that way ?

packages/jupyter-chat/src/tokens.ts
```diff
@@ -52,3 +52,39 @@ export const IChatPlaceholderFactory = new Token<IChatPlaceholderFactory>(
   '@jupyter/chat:IChatPlaceholderFactory',
   'The placeholder factory for the multi-chat panel'
 );
+
+/**
+ * The interface for the chat body placeholder factory.
+ */
+export interface IChatBodyPlaceholderFactory {
```
brichet, line 59: Currently `IChatBodyPlaceholderFactory` and `IWelcomeMessage` provide similar behavior (a placeholder when opening an empty chat). It may be confusing to keep both.
Perhaps, in a follow up PR, we could allow a `content?: string` in the props. If provided, this content could be rendered as markdown, instead of calling `create` (which would be optional too in this case).
That way, the `ChatBodyPlaceholder` could be use to both create a custom component or display a markdown, and we could remove the `IWelcomeMessage`.

brichet (commented):
Thanks @bhavana-nair and @dlqqq for working on this.

I just have a couple of comments below, otherwise it looks good to me.

packages/jupyter-chat/src/tokens.ts
dlqqq: Good callout, let me open a follow-up issue for this.

dlqqq: Done, see #436.

packages/jupyter-chat/src/components/chat.tsx
dlqqq: Good question. @bhavana-nair can you take a look and fix this?

bhavana-nair: Fixed this. Thanks for the review @brichet @dlqqq .

dlqqq (approved):
Thank you!

PR #434, opened 2026-05-11
[Bug] Fix chat auto-scroll on new messages and streaming content

Zsailer:
The chat wasn't reliably auto-scrolling to the bottom when new messages arrived, and had no way to re-engage scrolling mid-stream.

**Problems:**

1. The old implementation used CSS `overflow-anchor` to pin the scroll position. This doesn't work in Safari, and in Chrome/Firefox it caused a partial viewport shift when new messages arrived while scrolled up — not a full scroll, just an annoying jolt.

2. For streaming messages (where content grows over time), a one-shot scroll jumped to the start of the last message but didn't follow the content as it rendered.

3. If you scrolled up during a streaming message and then scrolled back down, there was no way to re-engage auto-scroll until the next message arrived.

**How it works now:**

- Removed CSS `overflow-anchor` entirely. The `ScrollContainer` is now a plain scrollable box with a forwarded ref — all scroll logic lives in `ChatMessages`.

- A `MutationObserver` watches the scroll container for DOM changes (new messages, streaming text, async markdown/code block rendering, images loading). When `shouldScrollRef` is true, it pins `scrollTop = scrollHeight` on every mutation.

- `shouldScrollRef` is controlled by two mechanisms:
  1. **On new message:** checks `model.messagesInViewport` to see if the previous last message was visible. If the user was at the bottom, pin. (Uses `length - 2` because `model.messages` already contains the new message by the time the signal fires.)
  2. **On scroll event:** if the user scrolls within 40px of the bottom, re-engage pinning. If they scroll up, disengage immediately — no waiting for the next message.

- Starts with `shouldScrollRef = true` so chat scrolls to bottom on first open.

**Tests** (`ui-tests/tests/autoscroll.spec.ts`):
- Chat starts scrolled to bottom on open
- Auto-scrolls to new messages when user is at bottom
- Viewport stays completely fixed when user has scrolled up
- Re-engages auto-scroll when user scrolls back to bottom

### DIFF
--- a/packages/jupyter-chat/src/components/messages/messages.tsx
+++ b/packages/jupyter-chat/src/components/messages/messages.tsx
@@ -84,7 +84,8 @@ export function ChatMessages(): JSX.Element {
     function handleChatEvents() {
       const viewport = model.messagesInViewport ?? [];
       const prevLastIdx = model.messages.length - 2;
-      shouldScrollRef.current = prevLastIdx < 0 || viewport.includes(prevLastIdx);
+      shouldScrollRef.current =
+        prevLastIdx < 0 || viewport.includes(prevLastIdx);
       setMessages([...model.messages]);
     }
     model.messagesUpdated.connect(handleChatEvents);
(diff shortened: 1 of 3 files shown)


brichet (approved):
Thanks @Zsailer, it looks great!

I think there is a linting error, otherwise it looks good to me.

Zsailer:
Thank you @brichet! :) 

Just push the linting fix.

Issue #436, opened 2026-05-15
Merge `IChatBodyPlaceholderFactory` and `IWelcomeMessage`


dlqqq:
Currently `IChatBodyPlaceholderFactory` and `IWelcomeMessage` provide similar behavior (a placeholder when opening an empty chat). It may be confusing to keep both.
Perhaps, in a follow up PR, we could allow a `content?: string` in the props. If provided, this content could be rendered as markdown, instead of calling `create` (which would be optional too in this case).
That way, the `ChatBodyPlaceholder` could be use to both create a custom component or display a markdown, and we could remove the `IWelcomeMessage`.

_Originally posted by @brichet in https://github.com/jupyterlab/jupyter-chat/pull/433#discussion_r3247465646_

dlqqq:
@brichet Can't a custom component already display Markdown? It seems like we can just drop `IWelcomeMessage`, no changes in `IChatBodyPlaceholderFactory` needed.

PR #437, opened 2026-05-15
Add configurable welcome message visibility (auto-hide and always-hide)

### LINKED ISSUE #224: Welcome message customization
### Problem

It would be great to have ways to customize the welcome message a bit more.

### Proposed Solution

Some ideas for options / settings that would be nice to have:

- ability to auto-hide the welcome message when a first message is sent (`true` by default)
- ability to always hide the welcome message (`false` by default), for those who don't want to see the message at all

### Additional context

The settings part may not need to be part of `jupyter-chat` but could be implemented by extensions, if `jupyter-chat` provides a way to hide the welcome message after it was added.

MUFFANUJ:
Closes #224 

This PR adds support for customizing welcome message visibility with two settings: `autoHideWelcomeMessage` (default true) to hide the welcome message after the first message, and `hideWelcomeMessage` (default false) to always hide it. The behavior is wired through core chat config and JupyterLab extension settings schema/loading.

### DIFF
--- a/packages/jupyter-chat/src/__tests__/model.spec.ts
+++ b/packages/jupyter-chat/src/__tests__/model.spec.ts
@@ -142,6 +142,8 @@ describe('test chat model', () => {
     it('should have empty config', () => {
       const model = new MockChatModel();
       expect(model.config.sendWithShiftEnter).toBeUndefined();
+      expect(model.config.autoHideWelcomeMessage).toBeTruthy();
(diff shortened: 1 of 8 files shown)


brichet:

For the reference, there is an issue wondering if we should keep the welcome message, or merge it with new placeholder: https://github.com/jupyterlab/jupyter-chat/issues/436

PR #439, opened 2026-05-25
Create the chat ID from frontend if not yet created

brichet:
This PR creates the chat ID from the frontend if there is no ID when loading the shared document.

The main idea is to allow using shared documents without server side collaboration (e.g. web RTC).


Zsailer (approved):
Makes sense! Looks good to me @brichet !

PR #442, opened 2026-06-06
Ensure chat files autosave independently

### LINKED ISSUE #430: Chat file may not be saved automatically
## Description

After https://github.com/jupyterlab/jupyter-collaboration/pull/479 (released in `jupyter-collaboration` [v4.1.0](https://github.com/jupyterlab/jupyter-collaboration/releases/tag/v4.1.0)), the collaborative documents may not be automatically saved if `autosave` is disabled in all clients.

## Reproduce

- install `jupyterlab-chat` with `jupyter-collaboration >=4.1.0`
- disable `autosave`
- write a message
- restart the server, the message is not in the chat anymore

## Expected behavior

The chat file to be automatically saved, independently of the `autosave` setting.

## Context

`jupyterlab-chat` is a special use case of the collaborative documents, and I wonder if we should:
- save the document anyway, whether `autosave` is disabled or not (make sense to me)
- add an indicator in the side panel that informs that the chat is `dirty`, and a way to save it

MUFFANUJ:
Closes #430 

Ensures collaborative chat files advertise autosave as enabled, preventing message loss when global document autosave is disabled. Adds focused regression tests for the awareness state behavior.


packages/jupyterlab-chat/src/autosave.ts
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+import type { IAwareness } from '@jupyter/ydoc';
+
+type AutosaveAwareness = Pick<
```
brichet, line 8: Just curious, is there any benefit of reducing the interface, or is it to simplify the tests writing ?
In any case it doesn't looks wrong, I don't have opinion on it.

brichet (approved):

That looks good to me, but maybe we should wait a bit to see if anyone else has any comments on this change, since it alters the expected behavior.

packages/jupyterlab-chat/src/autosave.ts
MUFFANUJ: It’s mainly to keep the helper’s contract narrow: it only needs `getLocalState` and `setLocalStateField`. The simpler tests are a nice side effect. Since it’s still a `Pick` from `IAwareness`, the method signatures stay tied to the upstream type.

brichet:
Let's merge this PR, it has been opened for 10 days now.

For the record, it forces the `autosave` state to `true` in the user awareness of the document (chat file). With this state to `true`, `jupyter-collaboration` automatically save the document on any change (https://github.com/jupyterlab/jupyter-collaboration/blob/c3ddf395b2faf1e1da2bb43a2eced803bd9e07e7/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L306)

PR #443, opened 2026-06-08
Faster playwright setup on CI

brichet:
Inspired by (and testing) https://github.com/jupyterlab/extension-template/pull/139


brichet:
Merging as it, the equivalent has already been merged in extension-template and Jupyterlab.

PR #444, opened 2026-06-08
Fix the UI tests on placeholder by improving the locator

brichet:
The locators used to select the chat from the placeholder was not precise enough, and landed on wrong chats created by others tests (e.g. in #419).

This PR fix it by expecting the chat name and not only the button in the placeholder.


brichet:
Merging it to fix flakiness on placeholder tests.

Issue #266, opened 2025-09-02
Better integrate with the JupyterLab toolbar factory machinery


jtpio:
### Problem

Currently, the Jupyter chat toolbar (next to the input text area) feels a bit different compared to every other toolbar in JupyterLab.

One of the differences is that the toolbar is directly next to the input chat area, which means that when the toolbar has many items, the chat area becomes cramped and the UX is degraded.

Another difference is that the chat toolbar is not configurable via the setting system, unlike the other toolbars, like the Notebook toolbar, or even the top area.

It would be great to check if we could integrate better with the base JupyterLab toolbar registry and factory mechanism? https://jupyterlab.readthedocs.io/en/latest/user/interface_customization.html#toolbars

At the same time, we could also add the ability to move the toolbar to the bottom to give the chat typing area a bit more space.

### Proposed Solution

Use the `IToolbarWidgetRegistry` to configure the chat toolbar.

And also take the opportunity to put the toolbar below the text area:

<img  height="500" alt="Image" src="https://github.com/user-attachments/assets/1872b25f-1cbf-4610-8604-789df8fcd40b" />

### Additional context

The `jupyterlab-chat` package makes use of the toolbar registry, but only for the toolbar at the top of the widget. So maybe we will need to move that logic to the main Jupyter chat package so it becomes available more generally.

<img height="600" alt="Image" src="https://github.com/user-attachments/assets/d0b91062-cceb-4f3c-b47f-70e13cc0a1fc" />

And in JupyterLab, many toolbars use the toolbar registry, such as the browser, notebook, and cell toolbars:

<img height="600" alt="Image" src="https://github.com/user-attachments/assets/6a7e510c-f2c3-4941-a68e-580a5b4d1f73" />

brichet:
The chat widget toolbar could also use some toolbar items defined from third party extensions.

For example #364 introduce a `delete` chat button, which make sense for `jupyterlab-chat` only. If we could define some toolbar item from `jupyterlab-chat`, they could be appended automatically to the widget.

Same in `jupyterlite-ai`, where several buttons are added after the widget creation, e.g. https://github.com/jupyterlite/ai/blob/d6a9c07fa818672e151dece65c04a1abf8ea82f3/src/index.ts#L564

PR #445, opened 2026-06-09
Configurable chat toolbar in side panel

brichet:
~~This PR adds a `chatToolbarItems` to the `MultiChatPanel`, to automatically add button to the toolbar when opening.~~

Adds a toolbar factory to the multichat panel, that is passed to the chat widget opened in the panel.
In `jupyterlab-chat`, the toolbar factory is built from the settings schema, and therefore have the same items as in the main area chat.

Some more context at https://github.com/jupyterlab/jupyter-chat/issues/266#issuecomment-4650361411

## API CHANGE

Since the same toolbar factory can be used in a main area chat and a chat in the side panel, this PR introduce a new interface to unify a chat panel (basically a panel containing a chat widget and a toolbar):
```ts
/**
 * The interface for any widget displaying a chat (main area or side panel).
 */
export interface IChatPanel extends Widget {
  /**
   * The chat widget embedded in the panel.
   */
  widget: ChatWidget;
  /**
   * The model of the chat widget.
   */
  model: IChatModel;
  /**
   * The area of the panel.
   */
  area: ChatArea;
  /**
   * The chat panel toolbar.
   */
  toolbar: Widget;
}
```

This widget interface is used to build the toolbar items in the registry, whether the widget is a main area chat or a side panel chat.

```ts
/**
 * An optional toolbar factory for each opened chat.
 */
 chatToolbarFactory?: (
   panel: IChatPanel
 ) => IObservableList<ToolbarRegistry.IToolbarItem>;
```

**This interface is also used as the tracked widget from the chat tracker (API change)**:
```diff
/**
 * the chat tracker type.
 */
- export type IChatTracker = IWidgetTracker<ChatWidget | MainAreaChat>;
+ export type IChatTracker = IWidgetTracker<IChatPanel>;
```

Any extension bumping to the next version of `@jupyter-chat` and providing an `IChatTracker` should update its widgets accordingly.


packages/jupyter-chat/src/utils.ts
```diff
@@ -81,3 +85,58 @@ export function replaceSpanToMention(content: string, user: IUser): string {
...
+ * This lets commands branch on `args.area` without each plugin having to
+ * register a per-item toolbarRegistry.addFactory call.
+ */
+export function injectAreaArg(
+  baseFactory: (
+    panel: IChatPanel
+  ) => IObservableList<ToolbarRegistry.IToolbarItem>,
+  commands: CommandRegistry
+): (panel: IChatPanel) => IObservableList<ToolbarRegistry.IToolbarItem> {
+  return (panel: IChatPanel) => {
```
dlqqq, line 101: I'm having trouble understanding exactly what this function's purpose is. It's only used in a single other area in `jupyterlab-chat-extension`. Why not just perform this logic inline where it is used?

packages/jupyter-chat/src/tokens.ts
```diff
@@ -3,24 +3,40 @@
...
 import { ChatWidget, Placeholder } from './widgets';
 import { IChatModel } from './model';
+import { ChatArea } from './types';
 
 /**
- * The main area chat widget type.
+ * The interface for any widget displaying a chat (main area or side panel).
  */
-export type MainAreaChat = MainAreaWidget<ChatWidget> & {
+export interface IChatPanel extends Widget {
```
dlqqq, line 17: Confirming my understanding: `IChatPanel` is an interface, and Jupyter Chat ships with 2 implementations:

- `MultiChatPanel` for the multi-chat view in the sidebar, provided by `@jupyter/chat`,
- `LabChatPanel` for the JupyterLab main area, provided by `jupyterlab-chat-extension`

This doesn't need to be addressed here, but if my understanding is correct, shouldn't both go into the same package? Is it because `DocumentWidget` is somehow specific to JupyterLab?

dlqqq (approved):
Looks like a great refactor overall, love the simplified interfaces. Thanks for working on this!

I left a couple of code questions below, but this looks generally safe. It doesn't look like there are any API changes that would affect most downstream consumers including Jupyter AI.

Does this require any JupyterLab v4.6+ APIs? Fine if it does, just wanted to check.

packages/jupyter-chat/src/utils.ts
brichet: You're right, it's only used in `jupyterlab-chat-extension` to add an the `area` arg in each command.
This arg is required to use the `move-chat` command, and I thought this workaround might also be useful for other extensions that provide a chat. It was mostly to avoid duplicating code, but I agree that this is confusing. Let's keep it in the extension requiring it.

packages/jupyter-chat/src/tokens.ts
brichet: Your understanding is correct. The idea behind the `IChatPanel` is to provide a single interface for a chat widget with a toolbar, that can be tracked by the `IChatTracker`.

IMO the `LabChatPanel` should remain in `jupyterlab-chat` because it is a `DocumentWidget`, which is really specific to `jupyterlab-chat`.
For example `jupyterlite-ai` also provides a main area chat, but it not a `DocumentWidget` https://github.com/jupyterlite/ai/blob/da63e32262d8c1a2305bde91910c07e09cfdd8af/packages/ai/src/widgets/main-area-chat.ts#L23

This interface should allow extension to provide any kind of widget including a chat, and to track it using the same widget tracker.

dlqqq (approved):

PR #447, opened 2026-06-16
Add "New Chat" entry to the file browser context menu

vincentye38:
## Summary

Add a `New Chat` entry to the file browser context menu (right-click). The entry appears alongside `New File` / `New Notebook` / `New Folder` from the JupyterLab core, at rank 53 so it sits naturally with the other "New ___" items.
Enhanced https://github.com/jupyterlab/jupyter-chat/pull/238

Implementation:
- Register a new command `jupyterlab-chat:create-in-folder` that:
  - If a folder is selected in the file browser, creates the chat inside that folder
  - Otherwise creates the chat in the file browser's current working directory
- Wire it declaratively via the `commands.json` schema (rather than `app.contextMenu.addItem`) so users can reorder/hide it through the JupyterLab settings UI like any other context-menu entry.

## Behavior

| Right-click target | Resulting chat location |
|---|---|
| Empty space in file browser | File browser CWD |
| A folder | Inside that folder |
| A file | File browser CWD (selected item isn't a directory, so fallback applies) |

The single command + single selector (`.jp-DirListing-content`) avoids menu duplication: Lumino renders only one "New Chat" entry per right-click, and the command resolves the right path at execution time based on the current selection.

https://github.com/user-attachments/assets/455b64bb-e0d5-4410-a205-323a18087a27

## Test plan

- [x] Right-click empty space in the file browser → "New Chat" appears between "New Notebook" and "New Folder"; selecting it creates an `untitledN.chat` in the current directory and opens it
- [x] Navigate into a subfolder, right-click empty space → chat is created in the subfolder
- [x] Right-click a folder → "New Chat" appears; selecting it creates the chat *inside* that folder
- [x] Right-click a file → "New Chat" appears; selecting it creates the chat in the file browser CWD (not next to the file)


packages/jupyterlab-chat-extension/src/index.ts
```diff
@@ -696,6 +696,20 @@ const chatCommands: JupyterFrontEndPlugin<void> = {
       }
     });
 
+    const createChatInFolder = 'jupyterlab-chat:create-in-folder';
+    commands.addCommand(createChatInFolder, {
```
brichet, line 700: Can you add the `describedBy` property ?
It should be minimal as there are no args, like in the command below (`markAsRead`).

brichet (commented):
Thanks @vincentye38, I just have a comment, otherwise it should be good to go.

packages/jupyterlab-chat-extension/src/index.ts
vincentye38: added describeBy at line #703

brichet (commented):
Thanks for updating quickly the PR.
Sorry for the second review, I forgot to ask if you could add tests on this menu entry ? Probably https://github.com/jupyterlab/jupyter-chat/blob/main/ui-tests/tests/commands.spec.ts is a good place for it.

vincentye38:
> Thanks for updating quickly the PR. Sorry for the second review, I forgot to ask if you could add tests on this menu entry ? Probably https://github.com/jupyterlab/jupyter-chat/blob/main/ui-tests/tests/commands.spec.ts is a good place for it.

Thank you for the reminder. Added tests.

ui-tests/tests/commands.spec.ts
```diff
@@ -119,6 +119,81 @@ test.describe('#launcher', () => {
...
 });
 
+test.describe('#contextMenu', () => {
+  test('should create a chat from file browser empty space', async ({
+    page,
+    tmpPath
+  }) => {
+    // Right-click on empty file browser area, then wait for the Lumino menu
+    // to render before clicking its items.
+    await page.locator('.jp-DirListing-content').click({ button: 'right' });
```
brichet:
```suggestion
    await page.locator('.jp-FileBrowser-Panel .jp-DirListing-content').click({ button: 'right' });
```

This selector seems to resolves to 2 elements sometime.
Specifying the parent element may help avoiding the test flakiness.

packages/jupyterlab-chat-extension/src/index.ts
```diff
@@ -696,6 +696,26 @@ const chatCommands: JupyterFrontEndPlugin<void> = {
...
```
krassowski, line 700: All other commands are referenced from `CommandIDs`. Should this one be added there too?

vincentye38: This is a private shortcut for the context menu to use only inside File Browser. I don't want to expose it to the public to call outside of File Browser. Is CommandIDs supposed to contains public commands only?
Please advise

brichet (approved):
Thanks @vincentye38, LGTM

PR #452, opened 2026-06-19
Compatibility to Jupyterlab 4.6.0

brichet:
This PR set the compatibility with jupyterlab >= 4.6.0
- use `jupyter-builder` to build the labextension and the python package
- bump `@jupyterlab/galata` in tests
- bump `jupyter-ydoc` and `pycrdt` dependencies


ui-tests/yarn.lock
krassowski, line 1: The additions here will decrease once we release https://github.com/jupyterlab/jupyterlab/pull/19054 that drops `jest-environment-jsdom` from `@jupyterlab/filebrowser` deps

krassowski (commented):
LGTM, thanks!

PR #455, opened 2026-06-30
Use partial message content to send message

### LINKED ISSUE #454: Replace the new message used in `sendMessage` by a partial message content
### Problem

The `IChatModel` interface in the frontend expect its implementation class to define `sendMessage()` method.
This `sendMessage()` receives an [`INewMessage`](https://github.com/jupyterlab/jupyter-chat/blob/064c3dfa1ef7ece8d1ad098a9cdde182d528aff8/packages/jupyter-chat/src/types.ts#L177) object (containing the body and an optional message ID), and is supposed to:
- build the full message:
  - add the message sender using the chat model user defined
  - add attachments and mentions using the `InputModel`
  - add timestamp
- propagate the message to other users

1. This seems not really logical, as the `sendMessage()` is almost used only in `InputModel`, so attachments and mentions could be built there.

2. It assume that there is only one user in the frontend, which prevent us to use frontend LLM agent to propagate messages.

3. It prevent sending a MIME bundle message from the frontend

### Proposed Solution

Instead of receiving an `INewMessage` as argument, the method could receive a partial [`IMessageContent`](https://github.com/jupyterlab/jupyter-chat/blob/064c3dfa1ef7ece8d1ad098a9cdde182d528aff8/packages/jupyter-chat/src/types.ts#L92).

That way, it would:
- allow using this method to send message from other frontend users 
- simplify the code for extensions implementing a chat model: no need to handle attachments and mentions, it should be handled by the input model and already formatted in the `Partial<IMessageContent>`
- allow sending MIME bundle message

### Additional context

Working on a frontend agent extension in https://github.com/jupyterlite/ai/pull/356, it can't used with `jupyterlab-chat` collaborative chat.
The messages from the frontend agent cannot be propagated to the shared model because of this limitation.

brichet:
This PR replaces the `INewMessage` interface (used to send message in the model) by a `Partial<IMessageContent>`.
It also handles attachments and mentions in the input model, instead of delegating it to the extension defining the `sendMessage()` method, as it was before the PR.

Fixes #454 (more context can be found in the issue).


packages/jupyter-chat/src/types.ts
```diff
@@ -174,10 +174,9 @@ export interface IChatHistory {
 /**
  * The content of a new message.
  */
-export interface INewMessage {
-  body: string;
-  id?: string;
-}
+export type INewMessage<T = IUser, U = IAttachment> = Partial<
+  IMessageContent<T, U>
+>;
```
dlqqq, line 182: `Partial<IMessageContent>` is too wide — it lets callers pass `id`/`time`/`type`/`deleted`/`edited`/`stacked`/`metadata`, none of which `sendMessage` respects (it assigns `id`/`time`/`type`/`raw_time` itself and drops the rest).

Only `body`, `sender`, `attachments`, `mentions`, `mime_model` seem like they are actually used. The type should probably be narrowed like this:

```ts
export type INewMessage<T = IUser, U = IAttachment> = Partial<
  Pick<IMessageContent<T, U>, 'body' | 'attachments' | 'mentions' | 'mime_model' | 'sender'>
>;
```

dlqqq (approved):
@brichet Thanks! New interface definitely looks cleaner. Left one nitpick about type-safety, but I understand if we are deliberately making `INewMessage` extremely general.

packages/jupyter-chat/src/types.ts
brichet: Good catch thanks, I updated it in https://github.com/jupyterlab/jupyter-chat/pull/455/commits/d2b8479744be730e1516741391bca30bdd8b9d63.
I also added the metadata attribute, and included it in *jupyterlab-chat* `sendMessage()` method.

brichet:
Test failure is not related to this change https://github.com/jupyter-server/jupyter_releaser/issues/659

dlqqq (approved):

PR #453, opened 2026-06-23
Bump @jupyterlab and @lumino dependencies

brichet:
Several packages from `@jupyterlab` namespace after `4.6.0` bumped `@jupyter/ydoc: ^4.0.0`, which includes API change (https://github.com/jupyter-server/jupyter_ydoc/pull/406), and therefore is not compatible with `@jupyterlab` packages prior to `4.6.0`.

This PR bumps `@jupyterlab` dependencies to `>=4.6.0`, and  prevents future conflicts if some dependencies are updated but other aren't, like below:

```bash
../../node_modules/@jupyterlab/notebook/lib/widgetfactory.d.ts:80:64 - error TS2344: Type 'NotebookPanel' does not satisfy the constraint 'IDocumentWidget<Widget, IModel>'.
  Types of property 'context' are incompatible.
    Type 'IContext<INotebookModel>' is not assignable to type 'IContext<IModel>'.
      Type 'INotebookModel' is not assignable to type 'IModel'.
        Types of property 'sharedModel' are incompatible.
          Property 'dirty' is missing in type 'ISharedNotebook' but required in type 'ISharedDocument'.

80     interface IFactory extends DocumentRegistry.IWidgetFactory<NotebookPanel, INotebookModel> {
                                                                  ~~~~~~~~~~~~~

  ../../node_modules/@jupyter/ydoc/lib/api.d.ts:87:5
    87     dirty: boolean;
           ~~~~~
    'dirty' is declared here.
```

PR #458, opened 2026-07-06
Clean the panel with toolbar workaround after bumping to jupyterlab>=4.6

brichet:
After https://github.com/jupyterlab/jupyter-chat/pull/453, including https://github.com/jupyterlab/jupyterlab/pull/18824, the panel with toolbar workaround added in https://github.com/jupyterlab/jupyter-chat/pull/415 can be safely removed.

cc. @MUFFANUJ


dlqqq:
I'm rebasing this PR because it seems like this depends on a JupyterLab v4.6 feature. The new CI workflows added in #463 will catch if this PR drops JupyterLab <4.6 support.

dlqqq:
Check release failure is not related -- I guess the tests wouldn't catch anything since we only do JupyterLab <4.6 testing for Python tests.

Still though, I'm not sure if this feature would work on JupyterLab <4.6 since the linked upstream PR was only released in JupyterLab 4.6.0 . If this is the case, that's fine, but we should document `jupyterlab-chat>=0.23.0` only supports `jupyterlab>=4.6`.

dlqqq:
Curiously, this fixes the E2E test failures on `main` reported in #464? Kicking CI to see if this is just flaky or if this really fixes them.

MUFFANUJ (approved):
Looks good. The JS package dependencies are already on JupyterLab `^4.6.0`; should we also update the remaining user-facing requirements/docs that still say `JupyterLab >= 4.0.0` so the supported minimum is consistent?

[python/jupyterlab-chat/README.md#L18](https://github.com/jupyterlab/jupyter-chat/blob/da2dc98e2706a536c38bc346afdf196ba850c169/python/jupyterlab-chat/README.md#L16-L18)

brichet:
I set this PR on draft until https://github.com/jupyterlab/jupyter-chat/issues/471 is fixed.

PR #462, opened 2026-07-07
Allow Jupyter Collaboration v5 alpha (Jupyter YDoc v4 support)

dlqqq:
## Summary

Bumps `jupyterlab-chat`'s dependency pins so it can install against the
**Jupyter Collaboration v5 alpha prerelease**, which is currently the only
Jupyter Collaboration release that supports **Jupyter YDoc v4**.

Jupyter Collaboration is a metapackage that ships several npm packages plus the
`jupyter_collaboration` Python package, and they don't all carry the same major
version. I verified the actual per-package versions at the `v5.0.0alpha0` tag /
published prereleases rather than assuming lockstep:

| Package | Version at v5 alpha |
|---|---|
| `jupyter_collaboration` (PyPI) | `5.0.0a0` |
| `@jupyter/collaborative-drive` (npm) | `5.0.0-alpha.0` |
| `@jupyter/docprovider` (npm) | `5.0.0-alpha.0` |
| `jupyter_ydoc` (transitive, via `jupyter-server-ydoc`) | `>=4.0.0b0,<5.0.0` |
| `@jupyter/ydoc` (transitive, via `collaborative-drive`) | `^4.0.0-a3` |

## Changes

- **`python/jupyterlab-chat/pyproject.toml`** —
  `jupyter_collaboration>=4,<5` → `jupyter_collaboration>=4,<6`. This keeps v4
  support while allowing the `5.0.0a0` prerelease. The existing `jupyter_ydoc`
  (`<5.0.0`) and `pycrdt` (`<0.15.0`) pins already satisfy what the alpha pulls
  in (`jupyter_ydoc 4.x`, `pycrdt 0.14.x`), so they're unchanged.
- **`packages/jupyterlab-chat/package.json`** and
  **`packages/jupyterlab-chat-extension/package.json`** —
  `@jupyter/collaborative-drive` `^4.4.0` → `^4.4.0 || ^5.0.0-alpha.0`. This is
  the one npm dependency jupyter-chat declares from the metapackage, and it
  reaches `5.0.0-alpha.0` at the v5 alpha. The `@jupyter/ydoc` range
  (`^3.0.0 || ^4.0.0`) already allows the v4 that the alpha depends on, so it's
  unchanged.
- **`yarn.lock`** — regenerated; `@jupyter/collaborative-drive` now resolves to
  `5.0.0-alpha.0` (with `@jupyter/ydoc 4.x`).

## Verification

- `pip install --pre "jupyter_collaboration>=5.0.0a0,<6"` resolves cleanly and
  installs `jupyter-collaboration 5.0.0a0`, `jupyter-ydoc 4.1.1`,
  `pycrdt 0.14.1`.
- Python test suite passes against the alpha (19 passed).
- Frontend builds successfully; `jlpm eslint:check` and `jlpm prettier:check`
  pass.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2400,15 +2400,15 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@jupyter/collaborative-drive@npm:^4.4.0":
-  version: 4.4.1
-  resolution: "@jupyter/collaborative-drive@npm:4.4.1"
+"@jupyter/collaborative-drive@npm:^4.4.0 || ^5.0.0-alpha.0":
+  version: 5.0.0-alpha.0
+  resolution: "@jupyter/collaborative-drive@npm:5.0.0-alpha.0"
   dependencies:
-    "@jupyter/ydoc": ^2.1.3 || ^3.0.0
-    "@jupyterlab/services": ^7.5.0
+    "@jupyter/ydoc": ^4.0.0-a3
+    "@jupyterlab/services": ^7.6.0-beta.1
     "@lumino/coreutils": ^2.2.1
     "@lumino/disposable": ^2.1.4
-  checksum: c4b05850e1f5791123d333993a8e18aa7aa96ce6a3cd85182603fbe4522fe795d77b8d523870a08ad21dfe5400b47684455b487bc7a7f2a44cb0f22e805f7c5c
+  checksum: 11b3f3d3a58d00b7794efa024f7bbe337198c65f1276896b71bb827b69a7845c1d1761962cf7b9f5105450534e8f663025d535db61c2ae7252a3aa23bf20b8aa
   languageName: node
   linkType: hard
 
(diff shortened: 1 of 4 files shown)


dlqqq:
@brichet I think this is safe because we want to support JupyterLab v4.6 anyways, which requires Jupyter YDoc v4, which is only released in Jupyter Collaboration v5 alpha right now. We can bump the version floors once the v5 official release is out.

I don't think this change will have an impact on most users. They will still install Jupyter Collaboration v4 if they run:

```
pip install jupyterlab jupyter-chat
```

Since without the `--pre` flag, `pip` will just install JupyterLab v4.6 but only Jupyter Collaboration v4. v4 doesn't support Jupyter YDoc v4 in its version range so their environments will still complain in the browser logs and `jupyter labextension list`. This will be fixed once the Jupyter Collaboration v5 official release is out.

I'd like to see this PR merged & released in Jupyter Chat v0.23.0a2 mainly to verify that we can get a Jupyter AI v3.1 pre-release working with no errors on JL v4.6 ASAP.

yarn.lock
```diff
@@ -2468,31 +2468,31 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@jupyter/ydoc@npm:^2.1.3 || ^3.0.0":
-  version: 3.0.5
-  resolution: "@jupyter/ydoc@npm:3.0.5"
+"@jupyter/ydoc@npm:^3.0.0 || ^4.0.0, @jupyter/ydoc@npm:^4.0.0":
+  version: 4.0.0
```
brichet: We could dedup dependencies, to keep only one version of each.
```bash
jlpm dlx yarn-berry-deduplicate -s fewerHighest && jlpm install
```

brichet (approved):
Thanks @dlqqq, LGTM.

I tested it with both set of versions of `jupyter-collaboration` and `jupyter-ydoc` successfully:
- jupyter-collaboration 4.4.1, jupyter-ydoc 3.5.0 and pycrdt 0.13.1
- jupyter-collaboration 5.0.0a0,  jupyter-ydoc 4.1.1 and pycrdt 0.14.1

yarn.lock
dlqqq: Good call — done in ee4f7d2. Ran `jlpm dlx yarn-berry-deduplicate -s fewerHighest && jlpm install`, which collapsed the duplicate `@jupyter/ydoc` 4.x resolutions (and a few `@jupyterlab/*` dupes) down to one version each (net −88 lines in `yarn.lock`). Rebuilt the labextension and re-ran lint — all clean.

dlqqq: Done, thank you!

PR #465, opened 2026-07-10
Add message metadata to the input model

dlqqq:
`INewMessage` already carries an optional `metadata` field, but the input model has no way to populate it, so it's always empty on send. We want the input toolbar to be able to stamp per-message metadata (for us, the target persona and its model/settings) onto the message it builds, without every consumer wiring its own path down to `send()`.

This adds a small `metadata` store to `IInputModel` alongside `mentions` and `attachments`: `getMetadata()`, `updateMetadata(patch)` (a lazy shallow merge), and `clearMetadata()`, plus a `metadataChanged` signal. `send()` now includes the metadata on the outgoing message when it's non-empty. Unlike attachments and mentions, the metadata is *not* cleared after sending — it holds a sticky selection that should carry to the next message until something changes it.

### DIFF
--- a/packages/jupyter-chat/jest.setup.js
+++ b/packages/jupyter-chat/jest.setup.js
@@ -14,3 +14,12 @@ global.IntersectionObserver = class IntersectionObserver {
   observe() {}
   unobserve() {}
 };
+
+/*
+ * `structuredClone` is a standard global in the browser (and Node), but the
+ * jsdom test environment doesn't expose it. Provide it so code that relies on
+ * it (e.g. input-model metadata, message cloning) can be unit-tested.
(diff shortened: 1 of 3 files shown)


dlqqq:
Part of the Jupyter AI v3.1 metadata-based routing work. Downstream PRs that consume this: jupyter-ai-contrib/jupyter-ai-persona-manager#59 (routes on `metadata.persona`) and jupyter-ai-contrib/jupyter-ai-acp-client#135 (the picker stamps the metadata). This lands first.

brichet (approved):
Thanks @dlqqq, it looks good.

PR #467, opened 2026-07-16
Expose awareness on IChatModel and IChatContext

andrii-i:
### Description

Extensions increasingly need the chat's Yjs awareness channel: jupyter-ai personas now publish their session state (models, settings, usage, commands) over it, and the chat input toolbar reads it. There is no public path to it today, so downstream code reaches through internals, e.g. jupyter-ai-acp-client reads `(chatContext as { _model?: unknown })._model.sharedModel.awareness` (https://github.com/jupyter-ai-contrib/jupyter-ai-acp-client/blob/73bd7aa/src/index.ts). Because the cast hides the access from the compiler, a rename of that internal breaks the consumer silently. Verified by renaming `_model` in a local build: everything still compiles and the consumer's slash-command E2E fails with completions never appearing.

This adds an optional readonly `awareness` to `IChatModel` and `IChatContext`. `AbstractChatContext` returns `this._model.awareness`, and `LabChatModel` implements it as `sharedModel.awareness`. Purely additive, no behavior change; `@jupyter/ydoc` is added to `@jupyter/chat` for the type (`^3.0.0 || ^4.0.0`, matching `jupyterlab-chat`).

Known consumers once released: jupyter-ai-acp-client replaces the reach-in, and the planned shared persona toolbar in jupyter-ai-persona-manager reads awareness through it from the start.

### Tests

Two unit tests in the core `model.spec.ts`: the context surfaces the model's awareness object, and it is undefined for models that don't provide one.

### DIFF
--- a/packages/jupyter-chat/src/model.ts
+++ b/packages/jupyter-chat/src/model.ts
@@ -9,6 +9,7 @@ import {
   nullTranslator,
   TranslationBundle
 } from '@jupyterlab/translation';
+import type { IAwareness } from '@jupyter/ydoc';
 import { ArrayExt } from '@lumino/algorithm';
 import { CommandRegistry } from '@lumino/commands';
 import { PromiseDelegate } from '@lumino/coreutils';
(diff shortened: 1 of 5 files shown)


dlqqq (approved):

packages/jupyter-chat/src/model.ts
```diff
@@ -53,6 +54,14 @@ export interface IChatModel extends IDisposable {
    */
   readonly ready: Promise<void>;
 
+  /**
+   * The awareness channel of the underlying shared document, when the chat is
+   * backed by one. Lets extensions read collaborative state attached to the
+   * chat (e.g. session information published by AI personas) through a
+   * supported API instead of reaching into a concrete implementation.
+   */
+  readonly awareness?: IAwareness;
```
brichet, line 63: @andrii-i why defining the awareness in `@jupyter/chat` ?
IMO, this is related to shared document only, so it should be defined only in the `LabChatModel`.

PR #470, opened 2026-07-21
[Bug] Fix chat rename dropping the file's directory

bsundaram1:
### Description

Renaming a chat via the "Rename chat" command drops the file's original directory. `renameChat`'s `execute` builds the new path directly from the bare name entered in the dialog (e.g. `test` → `test.chat`), so `Contents.rename("chats/untitled.chat", "test.chat")` moves the file to the workspace root instead of renaming it in place under `chats/`. This breaks any deployment that sets `defaultDirectory` to a non-root folder — every rename silently relocates the chat out of that directory.

This adds `resolveChatRenamePath(oldPath, newName)` to `jupyterlab-chat`'s `utils.ts`, which ensures the `.chat` extension and, when the new name has no explicit path segment, resolves it relative to `oldPath`'s directory instead of the workspace root. An explicit directory in the typed name (e.g. `archive/test`) is still respected as-is. `renameChat`'s `execute` in `jupyterlab-chat-extension` now calls this helper instead of only appending the extension.

Also fixes `jupyterlab-chat`'s `jest.config.js`: its local `transformIgnorePatterns` override had drifted out of sync with the `esModules` list shipped by `@jupyterlab/testing`'s base jest config, missing several ESM-only packages (`@jupyter/react-components`, `@jupyter/web-components`, `@marijn`, `@microsoft`, `color`, `exenv-es6`, `marked`) that transitively load when importing anything that touches `@jupyter/chat`. This surfaced only now because the new test is the first in this package to import a module that pulls in that chain.

### Tests

Added `src/__tests__/utils.spec.ts` with five unit tests for `resolveChatRenamePath`: keeps a bare rename in its original directory, ensures the `.chat` extension, leaves root-level chats at the root, respects an explicit directory in the new name, and preserves nested directories. Verified `jlpm jest` passes for the full `jupyterlab-chat` package (8/8), and `tsc` type-checks cleanly for both `jupyterlab-chat` and `jupyterlab-chat-extension`.

### Reviewer instructions

- Set `defaultDirectory` to a non-root folder (e.g. `chats`), create a chat, rename it via the "Rename chat" command using a bare name, and confirm the renamed file stays under `chats/` instead of moving to the workspace root.


packages/jupyterlab-chat/jest.config.js
```diff
@@ -7,9 +7,16 @@ const jestJupyterLab = require('@jupyterlab/testing/lib/jest-config');
...
+  '@marijn',
+  '@microsoft',
+  '@jupyter/react-components',
+  '@jupyter/web-components',
   '@jupyter/ydoc',
   '@jupyterlab/',
+  'color',
+  'exenv-es6',
   'lib0',
+  'marked',
```
dlqqq, line 19: Can these be removed? Why are these added?

dlqqq:
Thank you for making this fix! Just a small comment. Happy to cut a patch release after this.

packages/jupyterlab-chat/jest.config.js
bsundaram1: TL;DR I had to add these to get the tests running locally on my laptop. Let me try to explain my understanding below.

Jest's `transformIgnorePatterns` in `jest.config.js` seems to be a hand-maintained override of the `esModules` list that `@jupyterlab/testing`'s base `jest-config.js` already ships (`node_modules/@jupyterlab/testing/lib/jest-config.js`). The local override has drifted out of sync with that base list over time.

My new `utils.spec.ts` is the first test in this package to import something that pulls in the full `@jupyter/chat` barrel: 
```
utils.ts → token.ts → @jupyter/chat → @jupyterlab/application → @jupyterlab/apputils → @jupyterlab/ui-components
```

Without this change, running the new test throws
```
  SyntaxError: Unexpected token 'export'
  .../@jupyterlab/ui-components/node_modules/@jupyter/react-components/lib/index.js:1
```

`@marijn`, `@microsoft`, `color`, `exenv-es6`, and `marked` are additional ESM-only transitive deps in that same import chain that the base `@jupyterlab/testing` config already accounts for - I just brought this package's local list back in sync with it.

I've pushed a throwaway commit removing these additions to see if the failures happen on CI. I can push a follow-up commit restoring them if they fail on CI also.

bsundaram1:
@dlqqq This is the exact error I was seeing when I ran the tests locally, and it is fixed by adding back the jest config changes. Do you think there is a different way to address it?

brichet:
Thanks @bsundaram1 for fixing this issue, it looks good to me.

> Do you think there is a different way to address it?

I don't know if there is another way to fix, for me it looks good to add them back. These `transformIgnorePatterns` seem to often be there in `jest.config.js` files of jupyterlab extensions.

bsundaram1:
> Thanks @bsundaram1 for fixing this issue, it looks good to me.
> 
> > Do you think there is a different way to address it?
> 
> I don't know if there is another way to fix, for me it looks good to add them back. These `transformIgnorePatterns` seem to often be there in `jest.config.js` files of jupyterlab extensions.

Thanks for taking a look! I've reverted my test commit to add them back.

dlqqq (approved):
Makes sense. Thank you for fixing this bug! Would love to get this out in a patch release

dlqqq:
Think CI error is unrelated, will have to take a look tmrw to see how to fix it

bsundaram1:
> Think CI error is unrelated, will have to take a look tmrw to see how to fix it

I've added [6e09a3c](https://github.com/jupyterlab/jupyter-chat/pull/470/commits/6e09a3c91749cf568aabe0ec812ec1c346b687bc) which fixes this error on my laptop.

brichet (approved):

PR #472, opened 2026-07-29
Revert dependency bumps to support jupyterlab <4.6

brichet:
This PR reverts #453 and partially #452 

- revert `@jupyterlab` dependencies to `<4.6.0` in _yarn.lock_
- revert the use of `jupyter-builder` to build the extension, as it is compatible with `jupyterlab>=4.6` only

Fixes #471

Issue #471, opened 2026-07-29
Support JL<4.6 in latest stable Jupyter Chat?


dlqqq:
## Description

This PR bumped the minimum requirements of the `jupyterlab-chat` labextension to packages only available in JupyterLab v4.6+ : https://github.com/jupyterlab/jupyter-chat/pull/453

The problem is that now `jupyterlab-chat` requires `jupyter-collaboration`, but there is no stable release of that package that supports the new `jupyter_ydoc` v4 upgrade introduced in JL v4.6. This means that by default, almost all users get a broken environment when installing this:

- For users with JL <4.6 users, installing `jupyter-ai~=3.1` or `jupyterlab-chat~=0.23` results in this error when running `jupyter labextension list`:

```
$ jupyter labextension list
JupyterLab v4.5.10
/private/tmp/jai-compat/.venv/share/jupyter/labextensions
        jupyterlab-eventlistener-extension v0.4.0 enabled OK (python, jupyterlab_eventlistener)
        jupyterlab_pygments v0.3.0 enabled OK (python, jupyterlab_pygments)
        jupyterlab-notebook-awareness v0.1.0 enabled OK (python, jupyterlab_notebook_awareness)
        jupyterlab-chat-extension v0.23.0 enabled  X (python, jupyterlab_chat)
        jupyterlab-commands-toolkit v0.1.6 enabled OK (python, jupyterlab_commands_toolkit)
        @jupyter-ai/chat-commands v0.0.4 enabled OK (python, jupyter_ai_chat_commands)
        @jupyter-ai/persona-manager v0.1.2 enabled OK (python, jupyter_ai_persona_manager)
        @jupyter-ai/acp-client v0.2.1 enabled OK (python, jupyter_ai_acp_client)
        @jupyter-ai/router v0.0.5 enabled OK (python, jupyter_ai_router)
        @jupyter-ai-contrib/server-documents v0.3.1 enabled OK (python, jupyter_server_documents)
        @jupyter/collaboration-extension v4.4.1 enabled OK (python, jupyter-collaboration-ui)
        @jupyter/docprovider-extension v4.4.1 enabled OK (python, jupyter-docprovider)


"jupyterlab-chat-extension@0.23.0" is not compatible with the current JupyterLab
Conflicting Dependencies:
JupyterLab                  Extension       Package
>=4.5.10 <4.6.0             >=4.6.0 <5.0.0  @jupyterlab/application
[...]

PR #473, opened 2026-07-29
ci: run UI tests against JupyterLab <4.6 and >=4.6 via matrix

dlqqq:
## Problem

The UI-test suite only ran against a single JupyterLab version, so a regression that breaks JupyterLab <4.6 (see #471) would not be caught in CI.

## Change

Run the `jupyterlab` UI-tests job as a matrix across two JupyterLab lines, so the integration suite guards both:

| JupyterLab | jupyter_collaboration | Galata | Snapshots |
|---|---|---|---|
| `>=4.6,<4.7` | `>=5,<6` | `^5.6.0` | compared |
| `>=4.2,<4.6` | `~=4.0` | `~5.5.10` | ignored |

Getting the suite to pass reliably against both versions required fixing a few version-locking and test-isolation issues:

### Version pinning
- **Galata is version-locked to JupyterLab** (5.5.x → JL 4.5, 5.6.x → JL 4.6). Each matrix entry pins the matching Galata line via `jlpm add`.
- Downgrading Galata otherwise pulls a **second `@playwright/test`**, which breaks Playwright test collection. `ui-tests/package.json` pins playwright to a single version (`1.61.0`) via `resolutions` — it satisfies both Galata 5.5's `^1.57.0` and 5.6's `^1.60.0`.
- `jupyter_collaboration` 5.x is now released, so the `>=4.6` entry installs it as a normal stable release (no `--pre`).

### Single snapshot set
Reference snapshots render slightly differently across JupyterLab versions, so a shared snapshot set produces false pixel diffs. Rather than maintain a second, version-specific set of references, only the latest JupyterLab line owns snapshots (at the default top-level `…-snapshots/` location). The older line still runs the full suite for functional regressions but sets `JLAB_IGNORE_SNAPSHOTS`, which flips Playwright's `ignoreSnapshots` in `ui-tests/playwright.config.js` so screenshot comparisons are skipped there. The snapshot-update workflow regenerates the single (latest) set.

### Test isolation
Several tests assert absolute chat counts read from the chat list, which is populated from the chats in `defaultDirectory` (root by default). Because the UI tests share a single Jupyter server across parallel workers, chat files created by other test files leaked into these counts — flaky, and surfaced once the suite ran against JupyterLab <4.6. The count-sensitive suites in `placeholder.spec.ts` and `side-panel.spec.ts` now scope themselves to their own `defaultDirectory` via `mockSettings` and route uploads/queries through it.

## Updating snapshots

Comment **`please update snapshots`** on this PR to have the snapshot-update workflow regenerate the Linux references for the latest JupyterLab line.

This PR builds on #471 / #472 (JupyterLab <4.6 support), which is already merged to `main`.


dlqqq:
Great, this action run proves that the current v0.23.0 release fails on JupyterLab <4.6 : https://github.com/jupyterlab/jupyter-chat/actions/runs/30491550601/job/90710371003?pr=473

I'm going to rebase this onto #472 and see if the tests now pass. If that succeeds, then we can merge #472, rebase this PR again, and then merge this pending green CI - then we can publish a v0.23.1 patch release.

dlqqq:
bot please update snapshots

dlqqq:
bot please update snapshots

PR #475, opened 2026-07-30
Revert "Revert dependency bumps to support jupyterlab <4.6 (#472)"

krassowski:
This reverts #472 and bumps jupyter-builder to v1.2.0 pulling https://github.com/jupyterlab/jupyter-builder/pull/155

image


brichet:
Thanks @krassowski.

Indeed, everything works well on my side, but I wonder if we should wait before bumping `@jupyterlab` dependencies, to avoid confusing messages about extension not compatible:

```bash
$ jupyter labextension list
JupyterLab v4.5.10
/home/brichet/mambaforge/envs/test-chat-4.5/share/jupyter/labextensions
        jupyterlab-chat-extension v0.23.1 enabled  X (python, jupyterlab_chat)
        jupyterlab_pygments v0.3.0 enabled OK (python, jupyterlab_pygments)
        @jupyter/collaboration-extension v4.4.1 enabled OK (python, jupyter-collaboration-ui)
        @jupyter/docprovider-extension v4.4.1 enabled OK (python, jupyter-docprovider)


"jupyterlab-chat-extension@0.23.1" is not compatible with the current JupyterLab
Conflicting Dependencies:
JupyterLab                  Extension       Package
>=4.5.10 <4.6.0             >=4.6.0 <5.0.0  @jupyterlab/application
>=4.6.10 <4.7.0             >=4.7.0 <5.0.0  @jupyterlab/apputils
>=4.5.10 <4.6.0             >=4.6.0 <5.0.0  @jupyterlab/codemirror
>=6.5.10 <6.6.0             >=6.6.0 <7.0.0  @jupyterlab/coreutils
>=4.5.10 <4.6.0             >=4.6.0 <5.0.0  @jupyterlab/filebrowser
>=4.5.10 <4.6.0             >=4.6.0 <5.0.0  @jupyterlab/launcher
>=4.5.10 <4.6.0             >=4.6.0 <5.0.0  @jupyterlab/notebook
>=4.5.10 <4.6.0             >=4.6.0 <5.0.0  @jupyterlab/rendermime
>=7.5.10 <7.6.0             >=7.6.0 <8.0.0  @jupyterlab/services
>=4.5.10 <4.6.0             >=4.6.0 <5.0.0  @jupyterlab/settingregistry
>=4.5.10 <4.6.0             >=4.6.0 <5.0.0  @jupyterlab/translation
>=4.5.10 <4.6.0             >=4.6.0 <5.0.0  @jupyterlab/ui-components
```

Should we wait for jupyter-collaboration v5 finale release to have jupyterlab-chat fully compatible with jupyterlab 4.6 ?

krassowski:
I think these ranges could be simply changed in `package.json` file to avoid that warning? This is a declaration of compatibility that the extension owns (decides), if it is wrong then it should be corrected - but that's unrelated to jupyter-builder migration etc.

jupyter-collaboration v5: I plan to cut a release on Monday.

brichet:
> I think these ranges could be simply changed in package.json file to avoid that warning?

You mean keeping the current resolutions in `yarn.lock` but have `@jupyterlab` dependencies to `^4.5.0` in `package.json` ?
Sound nice, and would avoid false incompatibility messages.

krassowski:
Yes, that what I meant; alternatively, you could use jupyter-builder with v4.5.0 versions in lockfile, it will work regardless (but then it needs to be told to compile against 4.5.x).

brichet:
I think that it is fine to update the dependencies in lock file, except if it could raise incompatibilities with other extensions.
IIUC, the extension incompatibility message is only about the dependencies in `package.json`, and it's true that jupyterlab-chat is compatible with both `4.5` and `4.6`.

cc. @dlqqq if you have an opinion on it, since you first raised the incompatibility [issue](https://github.com/jupyterlab/jupyter-chat/issues/471).

brichet:
I added a commit to loosen the dependencies, and to remove the compatibility warning with `jupyterlab 4.5.x`.

### 4.5.10

```bash
$ jupyter labextension list
JupyterLab v4.5.10
/home/brichet/mambaforge/envs/test-chat-4.5/share/jupyter/labextensions
        jupyterlab_pygments v0.3.0 enabled OK (python, jupyterlab_pygments)
        @jupyter/collaboration-extension v4.4.2 enabled OK (python, jupyter-collaboration-ui)
        @jupyter/docprovider-extension v4.4.2 enabled OK (python, jupyter-docprovider)


Disabled extensions:
    @jupyterlab/filebrowser-extension:defaultFileBrowser
    @jupyterlab/notebook-extension:cell-executor
```

### 4.6.2

```bash
$ jupyter labextension list
JupyterLab v4.6.2
/home/brichet/mambaforge/envs/test-chat-4.6/share/jupyter/labextensions
        jupyterlab_pygments v0.3.0 enabled OK (python, jupyterlab_pygments)
        @jupyter/collaboration-extension v5.0.0 enabled OK (python, jupyter-collaboration-ui)
        @jupyter/docprovider-extension v5.0.0 enabled OK (python, jupyter-docprovider)


Disabled extensions:
    @jupyterlab/filebrowser-extension:defaultFileBrowser
    @jupyterlab/notebook-extension:cell-executor
```

brichet:
Failing tests are reproducible. It's may be related to the upgrade of `jupyter-collaboration 5`.

dlqqq:
Oof, I'm a bit short on time so might not get to this soon. But let's definitely squeeze this in before next minor release. Maybe best to get some pre-release wheels out to really confirm things work with JupyterLab v4.4 & v4.5.

brichet:
> confirm things work with JupyterLab v4.4 & v4.5

I did not tested it with jupyterlab 4.4 but there should be at least incompatibility warnings, even if it works.
We can probably live with it, as it should not be a breaking change for users to migrate to 4.5 if necessary.

brichet:
> But let's definitely squeeze this in before next minor release. 

I wonder if we should not release it in the next patch release `0.23.2`, since `0.23` was supposed to fix compatibility with jupyterlab 4.6 and jupyter-collaboration 5.

krassowski:
> Failing tests are reproducible. It's may be related to the upgrade of jupyter-collaboration 5.

It is. Specifically, this comes back to https://github.com/y-crdt/y-crdt/issues/632 which forced us to remove `self._document.dirty = False` from early initialization sequence (https://github.com/jupyterlab/jupyter-collaboration/pull/603/changes/59bd55ea942a9359015835b3e172b1f654b4b640). All this means is that the jupyter-chat still _works_ but the initialization ends up slower because the assumptions on the event sequence are no longer true.

To "fix" the tests we could wait until every chat currently attached to the shell has finished loading.

<details>

 A chat model resolves `ready` when its id becomes known, which happens either as soon as the document's `metadata.id` arrives, or - for a chat that does not carry one yet - only after the shared document reports a transition *out of* the dirty state, since that is what triggers generating an id. A fully synchronized document is therefore not necessarily a ready one.

 Until the model is ready, the input has no chat context, so `@`-mention completions resolve against an empty user list and dropped attachments are discarded. Neither the panel being visible nor the input container being present are usable signals: the input is in the DOM roughly 120ms after the chat is opened, whereas a chat without a stored id only becomes ready around 1.2s in.

`model.input.chatContext` is assigned exactly when `model.ready` resolves, which makes it a synchronously observable proxy for readiness that we could use in tests.

</details>

I am looking into a clean fix.

brichet:
Thanks @krassowski for the details.

Should we add an attribute to the DOM of the widget when the model is ready (to be able to check it in playwright) ?
Or do you think about something else ?

```ts
this.model.ready.then(() => {
  this.node.dataset.chatModelReady = 'true';
});
```

krassowski:
Something else - patching that in tests only does not remove 1-2 seconds delay which is user-facing. I have that documented and fixed locally, pushing them right now over in #478.

krassowski:
:green_circle:

brichet:
@dlqqq if it is fine for you too we can include it in a patch release.

brichet (approved):

brichet:
Looks like there are some conflicts after https://github.com/jupyterlab/jupyter-chat/pull/470, I'll rebase ASAP.

brichet:
> Maybe best to get some pre-release wheels out to really confirm things work with JupyterLab v4.4 & v4.5.

Tested with jupyterlab 4.4 as well, it works despite the warning about incompatibilities when running `jupyter labextension list`.

krassowski:
Yes, there is no reason it would not work. One approach forward is to merge this PR and update #473. Are we waiting on @dlqqq's approval or is it good to merge?

brichet:
Let's merge it and have a patch release with it.

PR #476, opened 2026-07-31
Fix the item positions in the reactive toolbar

brichet:
This PR fixes the item position in the reactive toolbar.

See https://github.com/jupyterlab/jupyterlab/issues/19226 for context.


dlqqq:
How far back are they willing to backport in JupyterLab right now? Ideally 4.4.x, since then no code changes here are needed.

dlqqq (commented):
If the upstream fix  can be backported to 4.4.x , I think that would be preferable. Otherwise, I'm fine with this as long as it has been tested :) 

Thank you!

PR #477, opened 2026-07-31
Return the sent message's id (or null) in `sendMessage` method

brichet:
Update the `sendMessage` abstract method to return either an string (id of the message sent) or `null` (if the message is has not been sent).

This would help handling streaming message in the frontend, to retrieve the initial message in the list.


dlqqq (approved):
This is technically a breaking change - do you want to wait on the next minor release for this?

Looks good to me otherwise.

brichet:
> This is technically a breaking change - do you want to wait on the next minor release for this?

Yes, we should wait for next minor release.
Let's fix compatibility with jupyterlab v4.6 and jupyter-collaboration v5 in the current version (v0.23), and move forward with this change after.

jtpio:
> Let's fix compatibility with jupyterlab v4.6 and jupyter-collaboration v5 in the current version (v0.23), and move forward with this change after.

Looks like the latest `0.23.2` release now addresses this? So maybe this PR can now be merged and released in a new `0.24.0` pre-release?

PR #484, opened 2026-08-11
Explicitly request file/directory content from the contents manager

brichet:
This PR specifies the query about the file/directory content when fetching it using the contents manager.

Related to https://github.com/jupyterlab/jupyterlab/pull/19081


dlqqq (approved):
Thanks, it's good to be as explicit as possible and guard against potential future changes.

PR #489, opened 2026-08-13
Make `jupyter-collaboration` optional

brichet:
This PR makes jupyter-collaboration an optional dependency.

See https://github.com/jupyterlab/jupyter-ai/pull/1642 for context.

### back end:

A common interface describes the chat model methods/attributes, and is implemented by `YChat` and `WsChatModel`.

When starting the application, it checks if RTC is available:
- one of `jupyter-server-ydoc` or `jupyter-server-documents` is enabled
- no trait disables those extension

The RTC state is published to `PageConfig`, and if RTC is disabled: 
- an handler is added to handle clients updates
- a `WsChatModel` is created for each opened chat file (path as reference)
- every update is propagated to all clients (user with file opened) and saved to the file

### front end:

When creating a chat model, it checks if RTC is enabled using `PageConfig` as source of truth.
If RTC is enabled, the `LabChatModel` propagates the new messages and updates using the shared document, as before this PR.
If RTC is disabled:
    - new messages and updates are propagated using a web socket
    - the server returns the new message/update to all clients (including the sender)
    - the shared model of the `LabChatModel` is updated on all client, which triggers a signal
    - the model of the chat widget is updated (from the signal), and the new message/update is displayed

### TODO
- [x] as suggested by @dlqqq in https://github.com/jupyterlab/jupyter-ai/pull/1642, we should have a robust way to check if RTC is available
- [x] as suggested by @dlqqq in https://github.com/jupyterlab/jupyter-ai/pull/1642, we should add a chat model in the backend, and have `YChat` and `WsChatModel` inherit from it (and make the chat discoverable)
- [x] allow the `WsChatModel` to be open in the main area


brichet:
I wonder if the front-end could be simplified by using just a single `LabChatModel` with its shared model.

We could add a `private _websocketHandler: WebSocketHandler | null` that would be `null` if RTC is enabled.

If RTC is disabled, the model would listen for changes on this `_websocketHandler` instead of the shared model, and would also propagate each change via the `_websocketHandler`.

And that would fix the document widget too.

dlqqq:
Sorry, didn't have much time today to work on this. Will still try to get to it.

@brichet It would be nice to get this ready to be merged though, and a clean `main` branch would be ideal for that. Can you help publish a v0.24.0 official release if it's ready? Then we can target this for v0.25.0 pre-releases, want to get something working available so I can start updating the Jupyter AI APIs to verify we have what we need.

dlqqq:
Rebasing onto latest main...

dlqqq:
Pushed the changes as described in the proposal to intelligently determine whether RTC is available. Implementation lives in `jupyterlab_chat.rtc_lib`. This struct gets published to `PageConfig`, and now sets the `collaborative` flag for the lab extension:

```ts
export interface IServerSessionRtcInfo {
  /**
   * Whether an RTC provider is active this session. When true, chat runs in
   * collaborative (Yjs) mode; when false, it uses the plain WebSocket
   * transport. This is the single source of truth for ``collaborative``.
   */
  enabled: boolean;

  /**
   * Which RTC backend is active (informational), or null.
   */
  provider: RTCProvider | null;

  /**
   * Diagnostic detail about RTC providers in this session.
   */
  providerDetails: {
    installed: string[];
    enabledByServer: string[];
    enabledByTrait: string[];
  };
}
```

I've included a matrix test running via `nox` that verifies this works correctly over all 3 environments (no RTC, RTC with `jupyter_collaboration`, RTC with `jupyter_server_documents`).

All of the basics work well without RTC:

Screenshot 2026-08-14 at 12 02 36 PM

dlqqq:
I've opened this follow-up issue to track other things I noticed but don't think need to block this PR: #490

dlqqq:
@brichet I think this is good to merge now. Added the configuration that allows for this to work seamlessly with or without `jupyter_collaboration`. I've cleaned up the `main` branch and the latest committed changes have all been released. I think this is good to merge, approving.

I wanted to squeeze in an E2E test suite with this PR, but these are taking a while and honestly I feel like this suffices as a proof-of-concept. We can do more in future PRs (like #497.

dlqqq (approved):

packages/jupyterlab-chat-extension/src/index.ts
```diff
@@ -125,36 +128,45 @@ async function createChatModel(
...
-    path: model.path,
-    format: model.format,
+  const modelFactory = app.docRegistry.getModelFactory(
+    'Chat'
+  ) as LabChatModelFactory;
+  const fileModel = await app.serviceManager.contents.get(path);
+
+  // The server's PageConfig decision is authoritative: only use the
+  // collaborative content provider when RTC is enabled this session.
+  const collaborative = getServerSessionRtcInfo().enabled;
```
brichet: We could use `isCollaborative()` directly, no ?

https://github.com/jupyterlab/jupyter-chat/pull/489/files#diff-152f8a38593edd35e87915c6f3a15e0af13e717e952c5bfda131e2365103f17dR84

packages/jupyterlab-chat-extension/src/index.ts
```diff
@@ -435,7 +447,26 @@ const docFactories: JupyterFrontEndPlugin<ChatWidgetFactory> = {
...
     app.docRegistry.addFileType(chatFileType);
 
-    if (drive) {
+    // The server decides whether RTC (collaborative) mode is active this
+    // session and publishes the decision via PageConfig. That decision is the
+    // single source of truth for `collaborative`: it honors which RTC server
+    // extensions are installed AND enabled, plus the jupyter_server_ydoc
+    // `disable_rtc` trait. The frontend `drive` provider is only the mechanism
+    // used when collaborative.
+    const collaborative = getServerSessionRtcInfo().enabled;
```
brichet: Same as above

packages/jupyterlab-chat-extension/src/index.ts
```diff
@@ -451,7 +482,11 @@ const docFactories: JupyterFrontEndPlugin<ChatWidgetFactory> = {
           translator,
           activeCellManager,
           selectionWatcher,
-          documentManager: filebrowser?.model.manager
+          documentManager: filebrowser?.model.manager,
+          collaborative,
+          serverSettings: collaborative
```
brichet, line 491: Should we check `collaborative && drive` here ?

brichet (commented):
Thanks @dlqqq for updating the PR.

For the record, we should also add i2c tests with both configuration.

brichet:
cc. @jtpio @krassowski @Zsailer if you have opinion on this change.

brichet:
@dlqq, another missing piece in this PR is a common way to retrieve the current chats opened, from the back end.
Currently in this PR, only the `WsChatModel` are available from the web app settings, under the key `"ws_chat_models"`. The `YChat` are available as before, by listening to an event signaling the creation of a new room.

We should make it generic (key `"chat-model"` registering `BaseChatModel`), and register also the `YChat` opened, as well as the web socket chat. Back end extensions would not need to know if RTC is available or not to use a chat model.

Should we handle it in this PR ?

dlqqq:
Gotcha. I will try getting E2E tests up and running and put together a list of E2E tests that don't work with RTC. Annoyingly GitHub is down ATM but hopefully this will be ready for review and merge by tmrw.

dlqqq:
@brichet 

Ahh, gotcha. Maybe what is needed is something like this, which lets consumers listen for filesystem updates without RTC: https://github.com/jupyter-ai-contrib/jupyterlab-live-content/

Maybe we need this in place:

```
/api/chats?dir="..."       # listen to updates on chat directories
/api/chat?path="..."       # WS for a single chat
```

dlqqq:
@brichet Oh I misunderstood what you meant. I see what you mean - we need to provide server extensions some way of knowing when a chat model was created or removed, which works without RTC.

The problem is that the existing RTC-based implementation's events are all handled by `jupyter_collaboration`, so either we build:

1. A generic API always emitted, which means duplicating some events on `jupyter_collaboration`

2. An RTC-free-specific events API that means consumers have 2 different ways to listen to chat events.

I am leaning towards the former (1).

brichet:
>A generic API always emitted, which means duplicating some events on `jupyter_collaboration`

I may be the way to go, triggering an event when a chat model is created or removed, with a new schema, e.g. `https://schema.jupyter.org/jupyterlab_chat/session/v1`.

But we'll also need a generic messages observer, currently for the `YChat` it is `ychat.ymessages.observe(callback)`, that uses the `ArrayEvent` object from `pycrdt`. To be fully generic, we should probably handle all objects (users, metadata and attachments).

It's probably for a follow up PR, as it may include a lot of additional changes.

dlqqq:
Have a draft implementation here: #499. Just emits chat open / closed events for now, I think it still needs a messages observer. Opened it as a separate PR, not going to let it block this one.

packages/jupyterlab-chat-extension/src/index.ts
dlqqq: Done in 542df5c — switched to `isCollaborative()`.

packages/jupyterlab-chat-extension/src/index.ts
dlqqq: Done in 542df5c — switched to `isCollaborative()`.

packages/jupyterlab-chat-extension/src/index.ts
dlqqq: Right now, I think we can keep it based on just what the server extension reports via `PageConfig`. If an admin enables RTC, we have to expect that the extension is configured correctly and RTC is enabled.

I don't think we call `drive` in many areas -- will have to double-check this (this is tracked in the follow up I opened before).

dlqqq: We already emit a console warning in case `!!drive !== collaborative`

dlqqq:
Addressed latest wave of feedback, E2E tests will be up shortly. Will merge this after as no other PR blockers were called out.

dlqqq:
We now run our JupyterLab E2E tests in a 2x2 matrix. JupyterLab <4.6 and >=4.6, RTC on and RTC off. All tests are passing.

Some of the tests are skipped when RTC is not available as they are not applicable. These tags are tagged with `collaborativeOnly`. 

- `#ychat` suite in `chat-file.spec.ts` as RTC does not use `YChat` or support out-of-band changes to chat files
- `#typingNotification` in `ui-config.spec.ts` as there is no need to emit client-side writing status in the non-RTC case
- `#user-mention` in `user-mention.spec.ts` similarly.


Screenshot 2026-08-17 at 2 01 25 PM

dlqqq:
CI is green, E2E tests are in place. All comments addressed. Proceeding to merge.

@brichet Thank you so much for all of your help with this PR!

PR #491, opened 2026-08-14
fix: keep message navigation working when deleted messages are hidden

### LINKED ISSUE #404: Message navigation is broken when deleted messages are hidden
## Description

When the property `showDeleted` is unchecked in settings and some messages are deleted in the chat, the message navigation is broken (cannot jump to last message anymore).

## Reproduce

- delete a message with the trash in the message toolbar
- uncheck the `showDeleted` option in settings
- reload the application
- the down arrow to jump to last message is not working anymore

## Expected behavior

The messages navigation to work.

## Context

The message navigation and unread messages use the message in viewport.
When the deleted messages are not shown, they are not rendered at all, an are never in the viewport.
Therefore, the messages are always unread and the navigation is broken.

uuzzrm:
Fixes #404

## What this does

The navigation buttons and the unread-message tracking work on indexes into the full model message list, but each rendered message was tagged with `data-index` equal to its position in the *filtered* list (deleted messages stay in the model, they just aren't rendered when `showDeleted` is off). As soon as one message is deleted, those two index spaces stop lining up:

- unread messages never appear in the viewport, so they never get marked as read and the navigation buttons stick around
- "go to last message" targets a DOM child that doesn't exist anymore, so the button silently does nothing

## How I fixed it

- `messages.tsx`: tag each rendered message with its index in `model.messages` instead of its position in the rendered list, so `data-index`, the viewport and the unread messages all live in the same index space.
- `navigation.tsx`: resolve navigation targets by `data-index` instead of `children.item(...)`, and use the last *rendered* message as the bottom target, so it still works when the last message in the model is deleted and hidden.
- `model.ts`: a deleted message is never rendered (or only as an empty placeholder) so it can never be marked as read by scrolling. Deleted messages are now excluded from the unread list when inserted, and dropped from it when a message is deleted, so a chat with a hidden deleted message doesn't stay permanently unread.

## Tests

- Two unit tests in `model.spec.ts` covering the unread behaviour for deleted messages.
- Three UI tests in `unread.spec.ts`: navigation to the last message works when a message in the middle is deleted and hidden, when the last message itself is deleted and hidden, and unread navigation still jumps past the hidden deleted message.

Verified locally: `tsc` build, the jest suite (54 tests), prettier and eslint all pass.

### DIFF
--- a/packages/jupyter-chat/src/components/messages/messages.tsx
+++ b/packages/jupyter-chat/src/components/messages/messages.tsx
@@ -210,6 +210,19 @@ export function ChatMessages(): JSX.Element {
     };
   }, [messages, showDeleted, allRendered]);
 
+  // The list of messages to display, once deleted messages are filtered out.
+  const visibleMessages = showDeleted
+    ? messages
+    : messages.filter(message => !message.deleted);
+
+  // Map each message to its index in 'model.messages'. The viewport and the
+  // unread messages are tracked against the full model list (deleted messages
+  // are kept in the model, even when they are not rendered), so 'data-index'
+  // has to reference the model index for the intersection observer and the
+  // navigation to stay consistent.
+  const modelIndex = new Map<string, number>();
+  model.messages.forEach((message, index) => modelIndex.set(message.id, index));
+
   const horizontalPadding = area === 'main' ? 8 : 4;
   return (
     <>
@@ -230,10 +243,7 @@ export function ChatMessages(): JSX.Element {
           className={clsx(MESSAGES_BOX_CLASS)}
         >
           {/* Filter the deleted message if user don't expect to see it. */}
-          {(showDeleted
-            ? messages
-            : messages.filter(message => !message.deleted)
-          ).map((message, i) => {
+          {visibleMessages.map((message, i) => {
             const isCurrentUser =
               model.user !== undefined &&
               model.user.username === message.sender.username;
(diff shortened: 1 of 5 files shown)


uuzzrm:
All checks are passing on the current head; the only red one is enforce-label, which requires a triage label on the PR (this one fixes a navigation bug when deleted messages are hidden, so "bug" seems right). Could a maintainer add the label so the check goes green? Thanks!

uuzzrm:
The UI test failures are resolved on the current head: both matrix jobs (JupyterLab >=4.6 and >=4.2,<4.6) now pass, and the build, typing, and Python unit suites are green too. The remaining enforce-label check needs a maintainer to add a triage label, which I don't have permission to do from a fork. Once that's applied the PR should be mergeable.

packages/jupyter-chat/src/components/messages/messages.tsx
```diff
@@ -210,6 +210,19 @@ export function ChatMessages(): JSX.Element {
     };
   }, [messages, showDeleted, allRendered]);
 
+  // The list of messages to display, once deleted messages are filtered out.
+  const visibleMessages = showDeleted
```
brichet, line 214: We should use `useMemo`, with `messages` and `showDeleted` as dependency here, to avoid computing it at each rendering.

packages/jupyter-chat/src/components/messages/messages.tsx
```diff
@@ -210,6 +210,19 @@ export function ChatMessages(): JSX.Element {
...
+    ? messages
+    : messages.filter(message => !message.deleted);
+
+  // Map each message to its index in 'model.messages'. The viewport and the
+  // unread messages are tracked against the full model list (deleted messages
+  // are kept in the model, even when they are not rendered), so 'data-index'
+  // has to reference the model index for the intersection observer and the
+  // navigation to stay consistent.
+  const modelIndex = new Map<string, number>();
```
brichet, line 223: Same here, the `modelIndex` could use `useMemo`

brichet (commented):
Thanks @uuzzrm, this looks good overall.

I just have some comments to avoid extra computation.

Issue #464, opened 2026-07-09
E2E tests flaky as of v0.23.0


dlqqq:
## Description

Check release is now failing -- this doesn't seem related, likely some upstream change that we need to pin in this step:

```
Ensuring mock GitHub
Collecting fastapi
  Using cached fastapi-0.139.0-py3-none-any.whl.metadata (26 kB)
Collecting uvicorn
  Using cached uvicorn-0.51.0-py3-none-any.whl.metadata (6.6 kB)
Collecting starlette>=0.46.0 (from fastapi)
  Using cached starlette-1.3.1-py3-none-any.whl.metadata (6.4 kB)
Collecting pydantic>=2.9.0 (from fastapi)
  Using cached pydantic-2.13.4-py3-none-any.whl.metadata (109 kB)
Requirement already satisfied: typing-extensions>=4.8.0 in /opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/site-packages (from fastapi) (4.16.0)
Collecting typing-inspection>=0.4.2 (from fastapi)
  Using cached typing_inspection-0.4.2-py3-none-any.whl.metadata (2.6 kB)
Collecting annotated-doc>=0.0.2 (from fastapi)
  Using cached annotated_doc-0.0.4-py3-none-any.whl.metadata (6.6 kB)
Requirement already satisfied: click>=7.0 in /opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/site-packages (from uvicorn) (8.1.8)
Collecting h11>=0.8 (from uvicorn)
  Using cached h11-0.16.0-py3-none-any.whl.metadata (8.3 kB)
Collecting annotated-types>=0.6.0 (from pydantic>=2.9.0->fastapi)
  Using cached annotated_types-0.7.0-py3-none-any.whl.metadata (15 kB)
Collecting pydantic-core==2.46.4 (from pydantic>=2.9.0->fastapi)
  Using cached pydantic_core-2.46.4-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (6.6 kB)
Collecting anyio<5,>=3.6.2 (from starlette>=0.46.0->fastapi)
  Using cached anyio-4.14.1-py3-none-any.whl.metadata (4.6 kB)
[...]

PR #492, opened 2026-08-14
test: make typing notification UI tests robust to slow CI

uuzzrm:
The three tests in the `#typingNotification` block are the flakiest in the suite right now (#464). Two of them fail for structural reasons, not because the feature is broken:

- `should display typing user` measured how fast the indicator appears with `Date.now()` around a Playwright assertion, and failed whenever the runner was slow. That bound is pure CI latency - it never actually verified what the comment claimed it verified (that the indicator disappears ~1s after typing stops).
- `should display several typing users` built a regex without capture groups and then read `result[1]`/`result[2]` from it, which are always `undefined`. Those two assertions were no-ops, so the test never confirmed that two *different* users show up.

What the changes do:

- `should display typing user`: wait for the indicator to appear, then wait for it to disappear. No wall-clock assertions. If the indicator ever sticks, the test fails instead of randomly timing out on a loaded runner.
- `should not display typing users if disabled`: same 3s window as before, but using `expect(...).not.toHaveText(/\S/)` instead of the manual `waitForCondition` loop.
- `should display several typing users`: the regex now uses a proper character class and the test actually asserts that two distinct usernames are listed.

Formatting is clean under `prettier --check`; the full UI suite runs in CI.

Refs #464.

### DIFF
--- a/ui-tests/tests/ui-config.spec.ts
+++ b/ui-tests/tests/ui-config.spec.ts
@@ -221,12 +221,13 @@ test.describe('#typingNotification', () => {
       .getByRole('combobox');
 
     await guestInput.press('a');
-    await expect(writers).toBeAttached();
-    const start = Date.now();
+
+    // The indicator should appear while the guest is typing...
     await expect(writers).toHaveText(/jovyan_2 is typing/);
 
-    // Message should disappear after 1s, but this delay include the awareness update.
-    expect(Date.now() - start).toBeLessThanOrEqual(2000);
+    // ...and disappear shortly after the guest stops typing (WRITING_DELAY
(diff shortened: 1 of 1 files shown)


uuzzrm:
All CI checks are green on the current head; the only red one is the enforce-label workflow, which needs a triage label on the PR. Since this hardens the flaky typing-notification UI tests, maintenance or ug both fit. Could a maintainer add one? Happy to adjust if a different label fits better.

uuzzrm:
All tests pass on the current head — the only red check is enforce-label, which needs a triage label from a maintainer. This one is a test-robustness change; “test” or “maintenance” would fit.

PR #493, opened 2026-08-15
fix: match @mentions with non-ASCII characters

uuzzrm:
## Description

The mention provider in `jupyterlab-chat-extension` parsed `@`-mentions with `/@([\w-]*)/g`, and `\w` is ASCII-only in JavaScript. Mention names keep the user's display name (see `IUser.mention_name`), so anyone with a non-ASCII name — `Pérez`, `Élodie`, `张三`, etc. — could not be completed or resolved: typing `@Pé` produced a truncated match, and on submit the mention was dropped entirely. The server-side `find_mentions` in `jupyterlab_chat/utils.py` uses Python's `\w`, which is Unicode-aware, so the client and server disagreed about what a mention name could be. This is the same issue reported for the Jupyter AI chat (jupyterlab/jupyter-ai#1393), which embeds this provider.

The fix moves the mention regex into `@jupyter/chat` as a shared `MENTION_REGEX` using a Unicode-aware character class (`\p{L}\p{N}_-`), and the provider now uses it. This matches what the server already accepts.

## Changes

- `packages/jupyter-chat/src/utils.ts`: export `MENTION_REGEX` (Unicode-aware, `g` flag preserved for `matchAll`).
- `packages/jupyterlab-chat-extension/src/chat-commands/providers/user-mention.tsx`: use `MENTION_REGEX` instead of the local ASCII regex.
- `packages/jupyter-chat/src/__tests__/mention.spec.ts`: unit tests covering ASCII, accented, hyphenated, non-Latin, multi-mention and punctuation cases (the accented cases fail against the old regex).
- `ui-tests/tests/user-mention.spec.ts`: UI regression tests for an accented display name (completion and mention attachment on submit).

## How to test

1. `cd packages/jupyter-chat && npx jest src/__tests__/mention.spec.ts` — 10 tests pass; the accented and non-Latin cases fail if the regex is reverted to `[\w-]`.
2. Full `@jupyter/chat` suite: 62 tests pass.
3. UI tests: `user-mention.spec.ts` includes `#non-ascii-mention` cases.

Refs jupyterlab/jupyter-ai#1393

### DIFF
--- a/packages/jupyter-chat/src/__tests__/mention.spec.ts
+++ b/packages/jupyter-chat/src/__tests__/mention.spec.ts
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+import { MENTION_REGEX } from '../utils';
+
+function mentionNames(text: string): string[] {
+  const names: string[] = [];
+  // exec() with a 'g' regex advances lastIndex, so build a fresh regex each
+  // call instead of reusing the shared one.
+  const regex = new RegExp(MENTION_REGEX.source, MENTION_REGEX.flags);
+  let match: RegExpExecArray | null;
+  while ((match = regex.exec(text)) !== null) {
+    names.push(match[1]);
+  }
+  return names;
+}
+
+describe('MENTION_REGEX', () => {
+  it('should match a plain ASCII mention', () => {
+    expect(mentionNames('@jovyan')).toEqual(['jovyan']);
+  });
+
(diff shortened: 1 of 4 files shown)


uuzzrm:
Checks are green except enforce-label. This is a small bug fix for @mentions with non-ASCII characters, so a “bug” label would let the check pass — could a maintainer add it?

brichet (commented):
Thanks @uuzzrm, this looks good overall.

brichet (dismissed):

brichet:
Test failures seem related, though.

PR #494, opened 2026-08-15
fix: don't send message when Enter confirms an IME candidate

### LINKED ISSUE #406: Chat input submits message when pressing Enter to confirm IME candidate
## Description

When using an Input Method Editor (IME) — such as Chinese Pinyin, Japanese, or Korean — pressing **Enter** to confirm a candidate character incorrectly submits the chat message instead of completing the IME composition.

The root cause is in the chat input `onKeyDown` handler, which does not check `event.isComposing` or `event.keyCode === 229` before treating Enter as a send action:

```javascript
// current code (simplified)
if ("Enter" === e.key && !m.menu.open) {
  // sends message immediately — even during IME composition
}
```

Suggested fix:

```javascript
if ("Enter" === e.key && !m.menu.open && !e.nativeEvent.isComposing && e.nativeEvent.keyCode !== 229) {
  // send message
}
```

References: [MDN: KeyboardEvent.isComposing](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing)

## Reproduce

1. Open a chat in JupyterLab with Jupyter AI installed
2. Switch to a CJK input method (e.g. Chinese Pinyin)
3. Type `jupyter` in the chat input — the IME shows a candidate list
4. Press **Enter** to confirm the selected candidate
5. The message is submitted immediately with incomplete/unintended text

Image

## Expected behavior

Pressing Enter during IME composition should confirm the selected candidate and insert it into the input field, not submit the message.

## Context

- Operating System and version: Linux (Kubernetes container), also reproducible on macOS
- Browser and version: Chrome
- JupyterLab version: 4.x
- jupyterlab_chat version: 0.21.1

<details><summary>Troubleshoot Output</summary>
<pre>
N/A — this is a frontend-only issue in the chat input component's keyboard event handling.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
N/A — no server-side errors involved.
</pre>
</details>

<details><summary>Browser Output</summary>
<pre>
No JavaScript errors in console. The issue is that the onKeyDown handler 
in the ChatInput component does not distinguish between a regular Enter 
keypress and an Enter keypress during IME composition.
</pre>
</details>

uuzzrm:
Closes #406

## Description

When typing with an Input Method Editor (IME) — Chinese Pinyin, Japanese, Korean — pressing **Enter** to confirm a candidate character was treated as a send: the message got submitted mid-composition with incomplete text. The keydown handler never looked at the composition state.

The fix checks `KeyboardEvent.isComposing` (and the legacy `keyCode === 229` marker some engines still emit) at the top of the input's `handleKeyDown` and bails out while a composition session is active, so Enter only sends a message when it's a real Enter press.

## Changes

- `packages/jupyter-chat/src/components/input/ime.ts`: small `isImeCompositionEvent()` helper with the composition check.
- `packages/jupyter-chat/src/components/input/chat-input.tsx`: `handleKeyDown` now returns early during IME composition.
- `packages/jupyter-chat/src/__tests__/ime.spec.ts`: unit tests covering active composition, the legacy keyCode marker, and plain key presses.

## How to test

- `jlpm workspace @jupyter/chat run test` (or `jest` inside `packages/jupyter-chat`)
- Manual: with a CJK input method active, type in the chat box, pick a candidate with Enter — the message should not send until composition is done.

All 8 suites / 66 tests in `@jupyter/chat` pass, and `prettier --check` + `eslint` are clean on the touched files.

### DIFF
--- a/packages/jupyter-chat/src/components/input/ime.ts
+++ b/packages/jupyter-chat/src/components/input/ime.ts
@@ -1,0 +1,19 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+/**
+ * Returns true when a key event is part of an ongoing IME composition
+ * (e.g. Chinese Pinyin, Japanese, or Korean input methods).
+ *
+ * Browsers set `isComposing` on key events emitted while a composition
+ * session is active, and some engines report `keyCode === 229` instead.
+ * Pressing Enter during composition confirms the current candidate; it
+ * must not be treated as a regular Enter press (e.g. sending a message).
+ */
+export function isImeCompositionEvent(
+  event: Pick<KeyboardEvent, 'isComposing' | 'keyCode'>
+): boolean {
+  return event.isComposing || event.keyCode === 229;
+}
--- a/packages/jupyter-chat/src/__tests__/ime.spec.ts
+++ b/packages/jupyter-chat/src/__tests__/ime.spec.ts
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+import { isImeCompositionEvent } from '../components/input/ime';
+
+describe('isImeCompositionEvent', () => {
+  it('should return true while a composition session is active', () => {
+    expect(isImeCompositionEvent({ isComposing: true, keyCode: 13 })).toBe(
+      true
+    );
+  });
+
+  it('should return true for the legacy keyCode 229 marker', () => {
+    expect(isImeCompositionEvent({ isComposing: false, keyCode: 229 })).toBe(
+      true
+    );
+  });
+
+  it('should return false for a plain Enter key press', () => {
+    expect(isImeCompositionEvent({ isComposing: false, keyCode: 13 })).toBe(
+      false
+    );
+  });
+
+  it('should return false for any other regular key press', () => {
+    expect(isImeCompositionEvent({ isComposing: false, keyCode: 65 })).toBe(
+      false
+    );
+  });
+});
--- a/packages/jupyter-chat/src/components/input/chat-input.tsx
+++ b/packages/jupyter-chat/src/components/input/chat-input.tsx
@@ -14,6 +14,7 @@ import {
 import clsx from 'clsx';
 import React, { useEffect, useRef, useState } from 'react';
 
+import { isImeCompositionEvent } from './ime';
 import { InputToolbarRegistry } from './toolbar-registry';
 import { submitInputMessage } from './submit-message';
 import { useChatCommands } from './use-chat-commands';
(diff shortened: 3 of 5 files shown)


uuzzrm:
Everything passes except the enforce-label check. This is a bug fix (Enter during IME composition shouldn’t send the message), so a “bug” label would unblock it.

dlqqq:
@uuzzrm Thank you so much for opening this wave of PRs. This bug especially has been a concern for Japanese users and I'm really grateful to see a fix proposed for this.

Note that there's a pretty major code change going on right now (#489). We can help rebase and merge these but note that it may be a bit before the PRs get merged. Will try to make sure this gets done by Friday.

brichet:
Duplicate of https://github.com/jupyterlab/jupyter-chat/pull/407

dlqqq:
@uuzzrm I'm inclined to merge this one since it's a little more robust (return early always on IME composition, instead of just if "Enter" is pressed), and this PR adds tests.

However, #407 by @aisensiy also adds this detection in `multichat-panel.tsx` , where a user may be searching for chats that contain CJK characters / requires IME composition. It would be great to add it there too.

Can you help port that specific change in #407 here, and credit @aisensiy in the commit message to give them proper attribution? Then I think this PR is good to merge.

dlqqq:
Also as a stretch goal, it would be great if we can have E2E test coverage for this, although I have no idea how that can be done.

uuzzrm:
Implemented the requested follow-up from #407.

- Added the IME-composition guard to `multichat-panel.tsx`, so confirming a CJK/IME candidate does not select a chat.
- Added a Galata/Playwright regression test covering both IME confirmation and regular Enter selection.
- Verified the regression test against the rebuilt labextension, including a negative check with the guard removed.
- Local unit tests: 56 passed.
- Side-panel E2E tests: 14 passed.
- TypeScript builds, Prettier, and targeted ESLint checks passed.

Credit: @aisensiy for the original `multichat-panel.tsx` change in #407.

This update was prepared with AI assistance and verified against the repository's test setup.

packages/jupyter-chat/src/components/input/ime.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Browsers set `isComposing` on key events emitted while a composition
+ * session is active, and some engines report `keyCode === 229` instead.
+ * Pressing Enter during composition confirms the current candidate; it
+ * must not be treated as a regular Enter press (e.g. sending a message).
+ */
+export function isImeCompositionEvent(
+  event: Pick<KeyboardEvent, 'isComposing' | 'keyCode'>
+): boolean {
+  return event.isComposing || event.keyCode === 229;
```
brichet, line 18: We should avoid use of `event.keyCode` as it is deprecated https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

brichet:
Curious to see how to reproduce the issue.
I tried using `ibus-pinyin` and `ibus-libpinyin`, pressing `enter` never send the message, instead it writes the original text, not the selected candidate, as it seems to be expected (`space` writes the candidate).
It acts the same with or without this change.

Issue #263, opened 2025-08-20
Allow extensions to control the status indicator


dlqqq:
### Problem

The new status indicator in Jupyter Chat looks delightful. I'm a big fan of how intuitive it is. There are a couple of features which would really enhance the AI extensions we are building on top of this:

1. **Allow chat users to set their own typing indicator message.** Right now, it always says "Jupyternaut is typing...". However, since Personas will be running agent tools, it would be nice if this were controllable by the Persona itself. For example, we would like to show "Jupyternaut is running `ripgrep /<some-regex>/ -- ./packages` ...".

2. **Move "stop streaming" button to status indicator.** Currently in Jupyter AI, we are using the configurable message footer to show a "Stop streaming" button at the bottom of each message. There are 2 issues:

- i) The stop button only appears if a message is being streamed. This is usually fine when just chatting. However, AI agents can run tools *before* streaming a natural language reply. If a persona hasn't streamed any content back yet and is running a tool that is taking too long, the user will not be able to halt it.

- ii) The stop button only appears under each message. If multiple personas are replying in parallel, it is difficult to stop all of them since a user may have to scroll to see different messages.

### Proposed Solution

1. Add an awareness API that allows the user to control the status indicator. Ideally it should support Markdown, but it is fine if the initial implementation doesn't.

2. Add an optional plugin that allows other extensions to add a custom React component into the status indicator, similar to the message footer previously added.

mfisher87:
> 1. **Allow chat users to set their own typing indicator message.** Right now, it always says "Jupyternaut is typing...". However, since Personas will be running agent tools, it would be nice if this were controllable by the Persona itself. For example, we would like to show "Jupyternaut is running `ripgrep /<some-regex>/ -- ./packages` ...".

I love this in general, but I think for a good UX the API should provide some implied guidance. For the `ripgrep` example, I want to see a rich and **persistent** tool call representation in the chat itself (https://github.com/jupyterlab/jupyter-ai/issues/1563#issuecomment-4374514557). A name in the API e.g. `actionVerb` might imply, but not require, that the developer should provide a short message for ephemeral display, ideally one word. E.g. `calling tools`, `loading`, `thinking`, etc.

PR #497, opened 2026-08-17
Support server-side writers without YDoc awareness

dlqqq:
## Motivation

In RTC-free (WebSocket) mode there is no awareness channel, so chat cannot show an "X is typing…" indicator the way it does under real-time collaboration. This matters most for AI personas: while a persona works, it should be able to signal that it is busy, ideally with its own text (e.g. "Jupyternaut is running ripgrep…") rather than a generic "is typing".

Jupyter Chat needs a transport-agnostic way for a server-side sender to advertise a writing status, so personas behave the same with or without RTC.

## Summary

This adds a writing-status API that lets the server show a typing indicator in RTC-free chat. Because that mode is effectively single-user, the client no longer advertises its own typing; instead the server (e.g. an AI persona) pushes writers to the client, which displays them. RTC mode is unchanged and still rides awareness.

The indicator text is customizable, so a sender can show what it is doing instead of a fixed "is typing" (addresses #263, part 1).

A sender starts and stops the indicator explicitly, and a stop takes effect immediately. As a safety net, a pushed indicator also auto-clears a few seconds after its last update, so a crashed or forgetful sender cannot leave it stuck.

## Technical details

- The shared chat model keeps writers in a per-user map and exposes a `broadcastWritingStatus` hook that each transport implements: RTC sets awareness; the WebSocket backend pushes an ephemeral writing frame that is never persisted.

- The backend `broadcast_writing_status(user, status)` takes the sender's own user, so agents with their own identity can advertise on their own behalf. Passing `status=None` stops.

## Testing

Unit tests for the writers state and indicator rendering on the frontend, and for the WebSocket writing relay on the backend.

### DIFF
--- a/packages/jupyterlab-chat/src/model.ts
+++ b/packages/jupyterlab-chat/src/model.ts
@@ -29,6 +29,15 @@ import { IChatChanges, IYmessage, YChat } from './ychat';
 const WRITING_DELAY = 1000;
 
 /**
+ * How long a server-pushed (e.g. AI persona) writing status stays visible
+ * without a refresh. The sender is expected to re-broadcast while still
+ * writing and to send an explicit stop when done; this is only a safety net so
+ * a crashed or forgetful sender cannot leave a "is writing" indicator stuck
+ * forever. An explicit stop clears it immediately, regardless of this value.
+ */
+const WS_WRITING_TIMEOUT = 3000;
+
+/**
  * Chat model namespace.
  */
 export namespace LabChatModel {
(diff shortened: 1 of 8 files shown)


dlqqq:
Actually this needs some equivalent of stopping awareness broadcast, right now it'll always show typing indicator for 2 seconds after AI persona stops writing

dlqqq:
K this is good now!

dlqqq:
Discussed with @brichet, we will review this together offline, but for now we will merge to allow others to start prototyping the changes in Jupyter AI.

packages/jupyterlab-chat/src/model.ts
```diff
@@ -375,22 +392,67 @@ export class LabChatModel
...
   };
 
-  private _resetWritingStatus() {
+  /**
+   * Broadcast the current user's writing status.
+   *
+   * Collaborative (RTC) mode advertises it over the awareness channel so peers
+   * see it. RTC-free (WebSocket) mode is effectively single-user, so the client
+   * does not advertise its own typing at all -- writers only ever come from the
+   * server (e.g. AI agents). Hence this is a no-op without RTC.
```
brichet, line 410: I think that this comment is incorrect, as websocket chat also work with multiple users AFAIK, even if this is not the main use case.

I'd like to clarify that I agree that there's no need to support broadcasting in the WS chat (at least for now).

PR #499, opened 2026-08-17
Server-side ChatManager and message observer

dlqqq:
## Motivation

Jupyter AI is a key dependent of Jupyter Chat. To power the AI experience it offers,  it needs to know two things about chats from the server: when chats open, close, or are deleted, and when to react to new messages. Both are currently wired to real-time collaboration internals.

- Chat creations are determined from events emitted from Jupyter Collaboration / JSD.

- Chat messages are observed from `ychat.ymessages` directly.

These are implemented in the `jupyter-ai-router` subpackage. To provide a RTC-optional experience, there needs to be generic ways to subscribe to both of these events without requiring RTC.

## Summary

This adds the server-side API the router needs, working the same way with or without RTC.

A `ChatManager` tracks the open chats in a generic registry keyed by path, emits lifecycle events (opened, closed, deleted), and hands out the chat model regardless of transport.

`observe_messages` on the shared chat model lets a consumer follow a chat's messages — received or edited by a client, sent or updated by the server — without reaching into the CRDT.

Together these let the router subscribe to chats and their messages transport-agnostically.

## Technical details

- The ChatManager registers on the web app settings (aliasing the existing `ws_chat_models` key), forwards jupyter-collaboration session events when RTC is present, and reclaims idle WebSocket models with a poller.

- The message observer wraps the shared array observer in RTC mode and emits directly from the WebSocket paths otherwise. In-place edits don't raise an array event in RTC, so only new-message events are surfaced there for now.

- This is the jupyter-chat side; the matching consumer changes live in jupyter-ai-router and are a follow-up.

## Testing

Unit tests for the ChatManager lifecycle and registry, and for the message observer on both transports.

### DIFF
--- a/python/jupyterlab-chat/jupyterlab_chat/events.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/events.py
@@ -265,9 +265,26 @@ class ChatManager(LoggingConfigurable):
         self._last_active[path] = time.time()
 
     def ws_client_gone(self, path: str) -> None:
-        # Do not free immediately; the inactivity poller reclaims it after the
-        # grace period unless a client reconnects.
+        # The last client for this chat has disconnected. Free the model now
+        # unless a server-side writer (e.g. an AI persona still producing a
+        # reply) is keeping it alive, in which case the poller reclaims it once
+        # the writer stops. Freeing here -- rather than reloading from disk on the
+        # next open -- is what keeps a reopened chat consistent without having to
+        # handle out-of-band file changes.
         self._last_active[path] = time.time()
+        if not self._has_active_writers(path):
+            self._free(path, ChatEventAction.CLOSED)
+
+    def _has_active_writers(self, path: str) -> bool:
+        """Whether a server-side writer is keeping this chat alive.
+
+        A "writer" is an AI persona (or other server-side producer) that is still
+        working on a reply and would be orphaned if the model were freed. The
+        server-side writing API is not on this branch yet (see
+        jupyterlab/jupyter-chat#497); until it lands there are no server-side
+        writers, so a chat is freed as soon as its last client disconnects.
+        """
+        return False
 
     def _get_or_create_ws(self, path: str) -> "WsChatModel":
         model = self._models.get(path)
@@ -277,6 +294,10 @@ class ChatManager(LoggingConfigurable):
             self._models[path] = model
             self._last_active[path] = time.time()
             self._emit_event(ChatEvent(path=path, action=ChatEventAction.OPENED))
+        # A cached model is the live in-memory session: reuse it verbatim. We do
+        # not reload from disk (out-of-band file changes are unsupported) so that
+        # any attached server-side state, such as AI personas, is preserved when a
+        # client reconnects.
         return model  # type: ignore[return-value]
 
     # ------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


brichet:
@dlqqq I missed time to review today, feel free to merge if it can help moving forward and testing, I'll take a look at it later.

dlqqq:
Discussed with @brichet, we will review this together offline, but for now we will merge to allow others to start prototyping the changes in Jupyter AI.

python/jupyterlab-chat/jupyterlab_chat/events.py
```diff
@@ -0,0 +1,363 @@
...
+        self._event_logger = self._settings.get("event_logger")
+        self._rtc_enabled = rtc_enabled
+
+        self._models: dict[str, "BaseChatModel"] = {}
+        self._room_to_path: dict[str, str] = {}
+        self._last_active: dict[str, float] = {}
+
+        # Single source of truth for the WS handler (supersedes the ad-hoc
+        # ``ws_chat_models`` dict; kept under the same key for compatibility).
+        self._settings["ws_chat_models"] = self._models
```
brichet, line 130: Do we need to keep a compatibility layer ?
The websocket handler is only available if there is no RTC, so `self._chat_manager.models` should be used as well.

python/jupyterlab-chat/jupyterlab_chat/events.py
```diff
@@ -0,0 +1,363 @@
...
+            if not (self._root_dir / path).exists():
+                self._free(path, ChatEventAction.DELETED)
+                continue
+            # Inactivity: only applies to WS models we own the memory for. A
+            # connected client keeps the chat alive.
+            if isinstance(model, WsChatModel):
+                if model.handlers:
+                    self._last_active[path] = now
+                elif now - self._last_active.get(path, now) > self.inactivity_timeout_s:
+                    self._free(path, ChatEventAction.CLOSED)
```
brichet, line 236: For clarification, is it to handle a server side client (LLM) still writing ?

My understanding is the following, please correct me if it wrong:
- if there is no "ws" client anymore, the memory should be freed by `ws_client_gone`
- if `model.handlers` is empty at that stage, it means that `self._has_active_writers(path)` is `True` (so the chat model has not been freed), we wait for `inactivity_timeout_s`

If this is the correct sequence, should we update the `_last_active(path)` when the server side client is writing (e.g. in `WsChatModel.add_message()` or `WsChatModel.update_message()` ?

python/jupyterlab-chat/jupyterlab_chat/websocket_handler.py
```diff
@@ -185,6 +189,11 @@ def _handle_new_message(self, data: dict, model: WsChatModel) -> None:
         model.broadcast(
             json.dumps({"type": "msg", "message": model.resolve_message(message)})
         )
+        received = model.get_message(message["id"])
+        if received is not None:
```
brichet, line 193: Should we broadcast the message to WS clients under the same condition (`received is not None`)?
Same for message update.

python/jupyterlab-chat/jupyterlab_chat/tests/test_events.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # second connection to same path: no duplicate `opened`
+        capture.clear()
+        mgr.ws_open("a.chat")
+        await _drain()
+        assert capture == []
+
+        mgr.stop()
+
+    asyncio.run(run())
```
brichet, line 63: Is there a reason for that pattern on these tests, instead of `async def test_...` ?

brichet (commented):
Thanks @dlqqq, this looks good overall.
I have some questions below.

PR #501, opened 2026-08-17
Add a transport-agnostic message observer API to the chat models

dlqqq:
## Motivation

Backend consumers such as the Jupyter AI router need to react to new chat messages. Today the only way to do that is to reach into `ychat.ymessages` and observe the shared CRDT array directly, which only works when real-time collaboration is enabled. Now that `jupyter-collaboration` is optional (#489), there is no way to listen to messages in the WebSocket configuration, so the router cannot function there.

## Summary

This adds a small message-observer API to the shared `BaseChatModel`: register a callback with `observe_messages` and release it with `unobserve_messages`, using the handle it returns so the subscription can't leak. Callbacks receive a typed event describing what happened to a message: received from or edited by a client, or sent/updated by the server. Both transports implement it, so a consumer can follow a chat's messages without knowing whether RTC is active.

## Technical details

- In RTC mode the model wraps the shared array observer under the hood and classifies each new message by its sender (a bot sender is treated as server, otherwise client). In-place edits mutate a nested map and don't raise an array event, so only new-message events are surfaced in RTC for now.

- In the WebSocket configuration all four event kinds are emitted directly from the message-handling paths.

- This is the model-side API only. Wiring the router to consume it transport-agnostically (via the server-side ChatManager) is a follow-up.

## Testing

Unit tests cover both transports: the event kinds, sender classification, and that unobserving stops delivery.

### DIFF
--- a/python/jupyterlab-chat/jupyterlab_chat/models.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/models.py
@@ -3,6 +3,7 @@
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
+from enum import Enum
 from typing import Any, Callable, Literal, Optional, Tuple, Union
 from jupyter_server.auth import User as JupyterUser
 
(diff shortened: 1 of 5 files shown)


dlqqq:
Folding into #499 — the ChatManager and this message-observer API are the same effort (Jupyter AI Router support for the RTC-free version), so they belong in one PR.

PR #503, opened 2026-08-18
Allow updating a message without setting the update state

brichet:
When a message in updated in the jupyterlab-chat model using the `updateMessage` method, the `edited` state is automatically set to `true`.
When a bot streams a message, it keep updating it, and therefore the message is seen as edited.
This PR allows calling the `updateMessage` method without setting the `edited` state.


dlqqq (commented):
1. Can't we just do this on whether the user is a bot? AFAIK this indication only needs to happen when a human edits a message.
2. This PR doesn't ever set `skipEditedUpdate` so there are no changes. Don't we want to set this to always be `false` always when AI is streaming a response via message updates?

**Some justification**

An experience where "edit a message" => "updates AI response" may not make sense for AI in Jupyter. Maybe we will allow user to edit the message while an AI response is in-progress, and the current AI response gets cancelled - but in that case there should be a notification like "user edited message, cancelling existing response and re-processing". Then new response should get appended on top. 🤔 

IMO the chat needs to serve as a record of everything the AI did and AI's should behave append-only, so it can't "erase its tracks" of what actions it took.

brichet:
Thanks @dlqqq for your feedbacks.
Indeed, it makes sense to avoid the flag for bot in any case.

I updated the PR, and added a test on it too.

dlqqq (approved):

PR #505, opened 2026-08-18
[0.24.x] Update the mime-model when updating message in jupyterlab-chat model

brichet:
This PR fixes a missing update of the mime-model when updating a message from `jupyterlab-chat` model.

~~This fix is not necessary in `main` branch, the message update as been updated and already fix it~~
Part of this PR is not necessary in `main` branch, the mime model is correctly updated to the YMessage
https://github.com/jupyterlab/jupyter-chat/blob/dd8dabe2cea3029059e68d3d2cae4befbeceefe6/packages/jupyterlab-chat/src/model.ts#L600-L602

but rerender needed to be fixed in `main` https://github.com/jupyterlab/jupyter-chat/pull/506

### DIFF
--- a/packages/jupyterlab-chat/src/__tests__/model.spec.ts
+++ b/packages/jupyterlab-chat/src/__tests__/model.spec.ts
@@ -1,0 +1,177 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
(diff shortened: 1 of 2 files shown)


dlqqq (approved):
LGTM. Leaving this open so you don't forget to cut a patch release of v0.24.x if needed :)

PR #506, opened 2026-08-18
Trigger a rerender of the message when the mime model only is updated

brichet:
When the mime model is updated in a message, the corresponding change in the `YMap` is not caught, and the message is not rerendered.

This PR fixes it.

Related to https://github.com/jupyterlab/jupyter-chat/pull/505


dlqqq (approved):
Thanks! Do consider just having a denylist though, since it seems it makes more sense to re-render by default on any change. Can probably find a more efficient way to do this via set intersections.

PR #507, opened 2026-08-18
Allow displaying cell output widget as message

brichet:
This PR add `jp-OutputArea` area class to the messages rendering a mime-bundle.

Some widgets are looking for this class to correctly display the widget.

More context in this [doc string](https://github.com/jupyterlite/ai/blob/93837d57d5b92576460f7f171d782a06640247ce/packages/ai/src/rendered-message-outputarea.ts#L7-L14) from `jupyterlite-ai`, included in https://github.com/jupyterlite/ai/pull/272.

cc. @jtpio

### DIFF
--- a/packages/jupyter-chat/src/components/messages/message-renderer.tsx
+++ b/packages/jupyter-chat/src/components/messages/message-renderer.tsx
@@ -20,6 +20,12 @@ const RENDERED_CLASS = 'jp-chat-rendered-message';
 const DEFAULT_MIME_TYPE = 'text/markdown';
 
 /**
+ * The class name added to an output area widget. This is required to display cell
+ * output as a message in the chat.
+ */
+const OUTPUT_AREA_CLASS = 'jp-OutputArea';
+
+/**
  * The type of the props for the MessageRenderer component.
  */
 type MessageRendererProps = {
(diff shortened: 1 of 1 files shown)


packages/jupyter-chat/src/components/messages/message-renderer.tsx
```diff
@@ -129,6 +137,8 @@ function MessageRendererBase(props: MessageRendererProps): JSX.Element {
       // This is necessary to render latex.
       MessageLoop.sendMessage(renderer, Widget.Msg.AfterAttach);
 
+      setIsOutputArea(!isMarkdownRenderer);
```
dlqqq, line 140: So the `.jp-OutputArea` class is always appended as long as the rendered content is not Markdown? Does this break anything?

dlqqq (approved):
LGTM, if this experience works well in JupyterLite and doesn't affect rendering negatively, I think this is safe too.

dlqqq:
I'll just merge this now since changes are safe. Not sure exactly what will happen but v0.25.0 will be pre-release for a while anyways, so let's test.

brichet:
@meeseeksdev please backport to 0.24.x

brichet:
> I'll just merge this now since changes are safe. Not sure exactly what will happen but v0.25.0 will be pre-release for a while anyways, so let's test.

Thanks @dlqqq, yes it should be safe.
I'll backport it to `0.24` patch release.

Issue #510, opened 2026-08-18
Pub/Sub API for Jupyter Chat


dlqqq:
## Proposal

Jupyter Chat can provide a generic pub/sub API that lets us do everything previously possible without RTC, while being compatible with both the RTC-free and RTC-based implementations. Consumers publish and subscribe to named topics; the server relays payloads to every subscriber of a topic. This one primitive can power many features that today each need bespoke wiring:

- new messages and message edits,
- typing state of web clients and of server-side agents,
- server-published session info in real time: available personas, models, and settings (see the persona-manager's [awareness models](https://github.com/jupyter-ai-contrib/jupyter-ai-persona-manager/blob/main/jupyter_ai_persona_manager/awareness_models.py)).

Under RTC, the YDoc and its awareness already provide this across the Y-protocol WebSocket. Awareness is especially versatile: any peer can attach arbitrary data under a key, and every other peer sees it, with no new handler code. That flexibility is why personas can broadcast their model and usage today without touching `jupyterlab-chat`.

The RTC-free path has no equivalent. `WsChatModel` ships a fixed protocol (`connection` / `users` / `msg` / `writing`) that only supports those message types. A consumer that wants to broadcast anything new must either patch `jupyterlab-chat` or open its own WebSocket per feature. The pub/sub API closes that gap: it gives the RTC-free path the same open-ended extensibility awareness gives the RTC path, behind one API that both transports implement.

## API

Add three methods to the base models: to `BaseChatModel` (per-chat channel) and to `ChatManager` (global channel):

```python
def pub(self, topic: str, data: Any) -> None: ...
def sub(self, topic: str, callback: Callable[[Payload], None]) -> SubToken: ...
def unsub(self, token: SubToken) -> None: ...
```

Everything on the wire is one payload:

```python
@dataclass
class Payload:
    client_id: str   # who published it (enables replicated sets + cleanup on leave)
    data: Any        # arbitrary; publishers and subscribers agree on the shape
```

A subscriber receives the current state of the topic on subscribe, then every subsequent payload. Because a WebSocket delivers in order and never drops a frame on a live connection, a client that joins mid-stream (e.g. while an agent is streaming a reply) cannot miss anything.

## Topics

Each topic is a category of payloads.

[...]

PR #511, opened 2026-08-19
[WIP] Generic pub/sub API for chat models (RTC + RTC-free)

dlqqq:
> [!WARNING]
> **WIP — this has not been reviewed by anyone yet.** Opened for visibility and CI. Design, naming, and scope are all subject to change. Do not merge.

Implements the generic pub/sub API proposed in #510, now on **both transports** (RTC and RTC-free).

## Core

- **`jupyterlab_chat/pubsub.py`**
  - `Payload { client_id, data }` — the single wire shape.
  - `PubSubBus` — in-memory relay: fans payloads to a topic's subscribers, retains the latest payload per publisher (catchup snapshot), and relays a clearing payload on retract/disconnect. It does not merge.
  - `MergeView` — the per-peer fold: dict payloads merge by top-level key (`null` deletes a key), non-dict replaces, `None` drops a publisher's contribution.
- **`BaseChatModel`** declares the API: `pub` / `sub` / `unsub` / `remove_client`, the reserved document-topic constants, and the `/chat/messages` bridge to `observe_messages`.
- **`ChatManager`** carries the same API for the global channel.

## RTC transport (`YChat`)

- **Document topics** map to the shared Y types: `/chat/messages` reuses the `observe_messages` bridge; `/chat/users`, `/chat/metadata`, `/chat/attachments` subscribe to the Y `Map`s (whole-map replace on change). `pub` routes to the typed methods (`add_message` / `set_user` / `set_metadata` / `set_attachment`).
- **Presence topics** map to awareness via **`awareness_pubsub.py::AwarenessBus`**. Each logical `client_id` owns its own awareness slot (swapping `awareness.client_id`, as the persona-manager does). The bus emits the *same raw-`Payload` stream* the in-memory bus does — one payload per contributing client, plus a `None` payload on retract/disconnect — so a consumer's `MergeView` folds identically on both transports. Awareness is created lazily over the `Doc` when the document isn't attached to a room.

## RTC-free transport (`WsChatModel`)

- **Document topics** route to the model's own state (persisted to `.chat`), with whole-map change notifications on `set_user` / `set_metadata` / `set_attachment`.
- **Presence topics** use the inherited in-memory `PubSubBus`; WS disconnect calls `remove_client` to drop a gone client's contributions (e.g. its writers entry).

## Testing

- `pytest python/jupyterlab-chat` → 93 passed, 2 skipped (existing + new `test_pubsub.py`, `test_pubsub_ychat.py`), covering the bus, `MergeView`, both transports' document + presence topics, and the global channel.
- `mypy python/jupyterlab-chat` → clean.

## Still out of scope

- The **frontend TypeScript** and the **WebSocket wire protocol** are not changed yet; this PR is the model-layer API on the server side. Wiring the browser client to `pub`/`sub` over the socket is the next step.

Tracks #510.

### DIFF
--- a/docs/design/pubsub-api.md
+++ b/docs/design/pubsub-api.md
@@ -118,7 +118,7 @@ to the same topic and stay in sync through the YDoc.
 
 The chat topic is special here: the server sends the message history the instant a
 client connects, because the client needs the file's contents to render anything.
-The web client therefore must be able to `sub("/chat/messages", ...)` *before* the
+The web client therefore must be able to `sub("/chat/messages", ...)` _before_ the
 WebSocket is open. Subscriptions are registered locally and flushed on connect, so
 the first thing the callback receives is the history catchup, followed by live
 edits on the same ordered connection.
--- a/docs/design/pubsub-api.md
+++ b/docs/design/pubsub-api.md
@@ -1,0 +1,189 @@
+# A pub/sub API for Jupyter Chat
+
+## Overview
+
+Jupyter Chat should expose a single, generic pub/sub bus. Consumers publish and
+subscribe to named topics; the server relays payloads to every subscriber of a
+topic. This one primitive can power many features that today each need bespoke
+wiring:
+
+- new messages and message edits,
+- typing state of web clients and of server-side agents,
+- server-published session info in real time: available personas, models, and
+  settings (see the persona-manager's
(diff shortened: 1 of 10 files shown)


dlqqq:
nah, let's not reinvent the wheel.

PR #513, opened 2026-08-19
Attribute server-side messages to a user via add_message(user=...)

### LINKED ISSUE #508: Allow server-side models to fake user info on send
### Problem

Multiple AI agents should be able to share the same chat, and for that to happen, they need to have different identities shown on their messages.

The current way we do this in Jupyter AI is to hack the awareness model to override the user that `YChat` sends from:

https://github.com/jupyter-ai-contrib/jupyter-ai-persona-manager/blob/main/jupyter_ai_persona_manager/persona_awareness.py

We also do a similar hack to allow a 'System' user to be shown in response to system-level commands (e.g. `/refresh-personas`):

https://github.com/jupyter-ai-contrib/jupyter-ai-persona-manager/blob/6adda36d188070a5b7cc510794af8db4f6510c1e/jupyter_ai_persona_manager/persona_manager.py#L391

### Proposed Solution

Provide a simple server-side Python-model-only API for adding users to the chat, and allow messages to be sent from there. Then, agents on the server can send messages attributed to one of these previously-added users:

```
BaseChatModel.add_user(user=...)
BaseChatModel.send_message(self, user=..., ...)
```

### Additional context

I am inclined to just call this "user" since you have to pass in a `User` model. But, maybe user is not the right name, open to suggestions here.

Long term, I think it would be easier to define an extremely simple `ChatUser` class that `add_user()` returns, e.g.

```
# persona calls add_user() to add itself to the chat
self.chat.add_user() -> ChatUser

# this ChatUser just forwards the method call to self.parent (BaseChatModel) and helps pass in the `user` argument

self.chat_user.send_message(...)
self.chat_user.broadcast_writing_status(...)
```

If folks are open to this I'm happy to include this with the PR.

dlqqq:
## Summary

Adds an optional `user` argument to `add_message`, so a server-side sender (such as an AI agent) can attribute a message to any identity it owns. When `user` is passed, it is registered in the chat if absent and the message is sent under that identity; otherwise the existing `new_message.sender` is used. This behaves identically on both transports (YChat and WsChatModel).

This replaces the awareness hacking that `jupyter-ai-persona-manager` does today to make server-side messages appear to come from a persona (or from a "System" user for system-level commands).

Also included is the small `add_user`/`ChatUser` convenience from the issue: `add_user(user)` registers a user and returns a `ChatUser` that remembers the model and identity, so an agent can call `send_message(...)` and `broadcast_writing_status(...)` without repeating its `user` on every call.

Closes #508

## Technical details

`NewMessage.sender` is now optional, since the sender can come from the `user` argument instead. `add_message` raises if neither a user nor a sender is given.

## Testing

New `test_chat_user.py` covers the `user` path on both transports (registration, sender attribution, sender override, the sender-or-user requirement, multiple bots sharing a chat) plus the `ChatUser` convenience layer and WebSocket writing-status forwarding.

### DIFF
--- a/python/jupyterlab-chat/jupyterlab_chat/__init__.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/__init__.py
@@ -15,7 +15,7 @@ import jupyter_ydoc  # noqa: F401
 
 from jupyter_server.utils import url_path_join
 
-from .models import BaseChatModel  # noqa: F401
+from .models import BaseChatModel, ChatUser  # noqa: F401
 from .rtc_lib import (  # noqa: F401
     RTC_PROVIDERS,
     RTCProvider,
--- a/python/jupyterlab-chat/jupyterlab_chat/models.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/models.py
@@ -364,3 +364,85 @@ class BaseChatModel(ABC):
         presence (e.g. the WebSocket model) override it.
         """
         # no-op by default
+
+    def add_user(self, user: "User") -> "ChatUser":
+        """Register ``user`` in the chat and return a :class:`ChatUser` bound to it.
+
+        This is the server-side entry point for adding a distinct identity (such
+        as an AI agent) to a chat. The returned :class:`ChatUser` remembers both
+        the model and the identity, so it can send messages and broadcast
+        writing status as ``user`` without passing the identity on every call.
+        """
+        self.set_user(user)
+        return ChatUser(model=self, user=user)
+
+    def send_message(
+        self,
+        user: "User",
+        body: str,
+        *,
+        mime_model: Optional[MimeModel] = None,
+        trigger_actions: list[Callable] | None = None,
+    ) -> str:
+        """Send a message to the chat attributed to ``user``.
+
+        ``user`` is registered in the chat first if it is not already present,
+        so a server-side sender can attribute a message to any identity it owns
+        without a separate :meth:`set_user`/:meth:`add_user` call. This lets
+        multiple AI agents share a chat while each sends under its own identity.
+
+        Returns the ID of the new message.
+        """
+        if user.username not in self.get_users():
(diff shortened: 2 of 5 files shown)


dlqqq:
@brichet @jtpio Curious what you two have to think about this - is there any reason to still have `get_user()` and `set_user()`? IMO the user should get passed explicitly to `chat.add_message()`, and if it is not set, it just shows a message from "System".

I think removing `get_user()` and `set_user()` makes sense if there is no use-case, and the new behavior would be fairly clear:

```py
chat.add_message() # send a System message as user is not specified

user = chat.add_user(...) # define a server-side user
user.send_message(...) # sends a message from this user, calls chat.add_message() with `user` arg
```

dlqqq:
I have more planned code changes which would clean this PR up and add more methods to the `ChatUser` interface, but I would like to first confirm this makes sense with you two.

brichet:
> I think removing get_user() and set_user() makes sense if there is no use-case

Not sure to understand your point @dlqqq, `set_user` method is still necessary to persist the user in the file, no ?
As far as I understand, the `add_user` method (added in this PR) uses the `set_user` and `get_user` methods. What is your plan to replace them if they get removed ?

dlqqq:
@brichet I intend for this PR to be purely additive, i.e. `chat.set_user()` still persists the user into the chat file, and `chat.get_user()` reads from that dictionary. I don't see any obvious reason to remove them.

The main thing about this PR is that saves server-side consumers from always having to pass the `user` argument everywhere. Server-side consumers would just call it like this:

```py
# in init
self.chat_user = self.chat.add_user( <identity> )

# usage
self.chat_user.add_message(...)
self.chat_user.broadcast_writing_status(...)
```

It's also easier to have a single object manage multiple users in the chat, e.g.

```py
self.admin_user = self.chat.add_user(<identity>)
self.system_user = self.chat.add_user(<identity>)
```

PR #514, opened 2026-08-19
Emit client connect/disconnect events

dlqqq:
Adds per-client connect/disconnect actions to the existing chat event bus.

## Why

The `jupyterlab_chat/room/v1` events (`opened`/`closed`/`deleted`) are **room-level**: `opened` fires once when a chat goes from no-clients to live, not per client. Server-side consumers (e.g. AI personas in `jupyter-ai-persona-manager`) need to know when *any* client connects so they can (re)publish per-client state — most importantly a **catch-up snapshot** (persona list, model/usage) to a client that joins an already-live chat. With RTC awareness being decoupled, there is otherwise no signal for "a new consumer connected."

## What

Rather than a separate schema, this extends the existing `room/v1` schema:

- Adds `client_connected` and `client_disconnected` to the `action` enum, plus an optional `client_id` field (present only for those actions).
- `ChatManager` gains `client_connected(path, client_id)` / `client_disconnected(path, client_id)` emitters; the same `observe_chats(...)` listener now carries these too (consumers filter by `action`).
- The WebSocket handler emits `client_connected` in `open()` and `client_disconnected` in `on_close()`, using the per-connection `client_id` it already assigns.

One event bus, one listener; purely additive to the existing `opened`/`closed`/`deleted` behavior.

## Testing

- New unit test `test_client_connected_and_disconnected_events` in `test_events.py`.
- Full Python suite: 69 passed, 2 skipped locally; `mypy` clean.

Note: emission is wired for the RTC-free WebSocket path (the decouple-RTC target). RTC per-client events can be forwarded from the collaboration provider in a follow-up if needed.

### DIFF
--- a/python/jupyterlab-chat/jupyterlab_chat/events.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/events.py
@@ -49,9 +49,9 @@ JUPYTER_COLLABORATION_EVENTS_URI = (
 CHAT_ROOM_EVENT_SCHEMA = {
     "$id": CHAT_ROOM_EVENT_SCHEMA_ID,
     "version": "1",
-    "title": "Chat room lifecycle events",
+    "title": "Chat room and client events",
     "personal-data": True,
-    "description": "Transport-agnostic chat room lifecycle events emitted by jupyterlab_chat.",
+    "description": "Transport-agnostic chat room lifecycle and per-client connection events emitted by jupyterlab_chat.",
     "type": "object",
     "required": ["path", "action"],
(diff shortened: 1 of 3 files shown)


dlqqq:
This change is purely additive and allows for events to get sent when a web client connects to the chat. Without this, other consumers have no way of knowing when a client connected to a chat. This was not needed before since YDocs auto-sync.

Since this is a purely additive change for the RTC-free model that's still in development, I am going to merge this. Hope folks are OK with that.

PR #515, opened 2026-08-19
Return typed attachments from `YChat.get_attachments()`

bhavana-nair:
The BaseChatModel abstract method declares:
  
```
  @abstractmethod
  def get_attachments(self) -> dict[str, Union[FileAttachment, NotebookAttachment]]:
```
  `YChat.get_attachments()` and `WsChatModel.get_attachments()` return different types
  **YChat.get_attachments() returned raw dicts violating the signature.**

Fixed so that all implementations of BaseChatModel.get_attachments() return the declared type.

### DIFF
--- a/python/jupyterlab-chat/jupyterlab_chat/ychat.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/ychat.py
@@ -215,19 +215,12 @@ class YChat(YBaseDoc, BaseChatModel):
         Values are typed as `FileAttachment` or `NotebookAttachment`
         """
         result: dict[str, Union[FileAttachment, NotebookAttachment]] = {}
-        for att_id, att_dict in self._get_attachments().items():
+        for att_id, att_dict in (self._yattachments.to_py() or {}).items():
             if att_dict.get("type") == "notebook":
                 result[att_id] = NotebookAttachment(**att_dict)
             else:
                 result[att_id] = FileAttachment(**att_dict)
         return result
-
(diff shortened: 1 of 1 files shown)


python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -211,6 +211,21 @@ def get_attachments(self) -> dict[str, Union[FileAttachment, NotebookAttachment]
...
+            else:
+                result[att_id] = FileAttachment(**att_dict)
+        return result
+
+    def _get_attachments(self) -> dict[str, dict]:
+        """
+        Returns the attachments of the document as raw dicts. Used internally
+        where a JSON-serializable form is required.
         """
         return self._yattachments.to_py() or {}
```
dlqqq: Do we still need this? Can we just put this in `get_attachments()`?

Wherever we need a proper dictionary, we can replace it with `asdict(self.get_attachments())`.

dlqqq (commented):
Wow, good catch!

python/jupyterlab-chat/jupyterlab_chat/ychat.py
bhavana-nair: Updated to use get_attachments()

dlqqq (approved):

dlqqq:
This looks like a clear bug fix, and I think this is safe to merge. Merging.

PR #516, opened 2026-08-20
Implement `YChat.broadcast_writing_status()`

dlqqq:
## Motivation

The Writers API (#497) let server-side senders such as AI personas drive the chat typing indicator, but it was only wired up for the RTC-free WebSocket model. In collaborative mode the `YChat` model inherited a no-op, so a persona writing a reply never actually showed as writing, and the stop button that keys off an active writer stayed dead.

This went unnoticed because consumers used to write the persona's awareness slot themselves. Once they stop doing that, the writers API turns out to have never been complete for the RTC path.

## Summary

Server-side writers now show as writing in collaborative mode, same as over WebSocket. `broadcast_writing_status` is also now abstract on the base model, so every transport has to implement it and a future backend cannot silently regress to a no-op.

## Technical details

Rather than fake an awareness client per writer, `YChat` publishes the whole set of current writers as a list under its own awareness slot; clients scan every slot for that field. The awareness channel already keeps a participant's own slot alive, so there is no heartbeat to manage. Writing status is ephemeral and never persisted to the `.chat` document.

On the frontend, the collaborative model now merges that published set with a peer's own typing slot into the existing `writersChanged` signal, so every component reads writers from one transport-neutral source whether or not RTC is on. This is a slightly breaking change to how writers are represented in awareness.

## Testing

Python tests assert the writer set published on the awareness slot (message-id, typing-only, stop, concurrent writers, non-persistence, and the no-room no-op). A frontend unit test covers the awareness-to-writers merge (both shapes, dedupe, local-user exclusion).

### DIFF
--- a/packages/jupyterlab-chat/src/__tests__/writers-awareness.spec.ts
+++ b/packages/jupyterlab-chat/src/__tests__/writers-awareness.spec.ts
@@ -1,0 +1,179 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+import { Doc } from 'yjs';
+import {
+  applyAwarenessUpdate,
+  Awareness,
+  encodeAwarenessUpdate
+} from 'y-protocols/awareness';
+
+import { collectWritersFromAwareness } from '../model';
+
+// The current user, occupying the local awareness slot. `IChatModel.name`-style
+// identity: writers are matched/excluded by `user.username`.
+const ME = { username: 'me', name: 'Me', display_name: 'Me' };
+
+function makeAwareness(): Awareness {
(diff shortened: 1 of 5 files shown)


dlqqq:
Reviewed everything line-by-line. These changes only affect the RTC-version and restore its existing functionality.

Basically: the server can advertise the list of server-side writers on its local state (i.e. personas), and each web client still just sets "isWriting" and "typingIndicator" on their local states. In the frontend, to get all the writers, we just iterate over all of the local states and combine the server-side writers + web client writers, while excluding self.

The tests use a real Awareness object to test as closely as possible, and assert that the client itself is always excluded from the writers list.

Since this change just restores a feature that was broken before, proceeding to merge.

PR #518, opened 2026-08-20
Require `User` to be a dataclass in backend

dlqqq:
## Summary

`broadcast_writing_status` hand-enumerated the writer's `User` fields when building the WS frame, which silently dropped `bot`. RTC-free clients then received a `writing` status whose `user.bot` was `undefined`.

Any consumer that distinguishes AI writers from humans via `user.bot` breaks: e.g. a stop/cancel button that enables only while an AI persona is writing (`chatModel.writers.some(w => w.user.bot)`) never enables, even though the backend `User(bot=True)` set it and the frontend (`IUser.bot`, `_onWsWriting`) passes the user through verbatim.

## Fix

`user` is always a `User` (a dataclass); no caller passed a dict. So:

- Require `user: User` (dropped the unused `Union[User, dict]` path).
- Serialize with `dataclasses.asdict(user)` instead of a hand-written field list, so **every** field -- including `bot`, and any added later -- survives the round trip. This removes the drift that caused the bug.

Adds a regression test asserting `bot` is preserved for both a bot and a human writer.

## Testing

`pytest jupyterlab_chat/tests` -- 68 passed, 2 skipped.

## Context

Surfaced by the jupyter-ai-persona-manager RTC-removal E2E matrix: the RTC-free (`WsChatModel`) leg's cancel-response test failed because the persona stop button never enabled. This fixes that transport. The RTC (`YChat`) transport has a separate gap -- `broadcast_writing_status` is still a no-op there -- tracked separately.

### DIFF
--- a/python/jupyterlab-chat/jupyterlab_chat/models.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/models.py
@@ -349,16 +349,15 @@ class BaseChatModel(ABC):
 
     def broadcast_writing_status(
         self,
-        user: Union["User", dict],
+        user: "User",
         status: Optional[dict] = None,
     ) -> None:
         """Broadcast an ephemeral "user is writing" status on behalf of ``user``.
 
-        ``user`` is a :class:`User` (or a mapping with at least ``username``),
-        allowing server-side senders such as AI agents -- which each have their
-        own user identity -- to advertise a typing indicator. ``status`` is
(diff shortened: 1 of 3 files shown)


dlqqq:
This one is pretty straight-forward, just fixes a dumb oversight introduced by AI in #497. The `User` model has been well-defined, methods always return it, so there's no reason to not just require a `User` instance here.

Merging pending green CI, hope that is OK with folks. This will help PR #516 pick up this change.

PR #519, opened 2026-08-20
Announce the writing indicator to assistive technology

opensource-joe:
The writing indicator already says something useful. It reads "Alice is typing...", or when a writer supplies its own phrase, "Jupyternaut is running `ripgrep`". None of it reaches a screen reader, because the container is not a live region. So the only signal that a reply is on its way is visible text that is never spoken.

There are no live regions anywhere in the package today: no `aria-live`, `role="status"`, `role="log"`, `role="alert"` or `aria-atomic` across all 109 `.ts`/`.tsx` source files. This adds the one that matters most.

## The change

Three attributes on the existing container, no visual change and no new dependency.

**`role="status"` and `aria-live="polite"`.** Polite rather than assertive, because a reply arriving is not urgent enough to interrupt whatever the reader is part-way through hearing.

**On the container rather than the text**, so the region is in the accessibility tree before anyone starts writing. A live region created at the same moment its content appears is unreliable: some assistive technology only announces changes to regions it was already tracking.

**`aria-atomic="true"`**, so the phrase is read as a whole. Without it a change to part of the text can be announced on its own, and a reader hears a bare name with no indication of what it refers to. That is a real risk here, because the text is rebuilt as writers join and leave.

The placeholder is unaffected. The container still reserves its 16px, and the non-breaking space it holds when nobody is writing stays out of the accessibility tree, because the text is `visibility: hidden` in that state.

## Tests

Added `writing-indicator.spec.tsx`, covering the live-region attributes, the announced phrase in both the default and custom-indicator forms, and that the placeholder is not announced as if it were a message.

It uses React's own `act` and `createRoot` rather than a testing library, so **no new dependency**. I checked the two accessibility assertions genuinely fail without the change:

```
✕ is a polite live region, so a reply on its way is announced
    Expected: "status"   Received: null
✕ reads the whole phrase rather than a fragment of it
    Expected: "true"     Received: null
```

The other three pass either way, which is intentional: they are regression guards on behaviour this change must not alter.

Full package suite is green, **8 suites, 62 tests**. `build:lib` typechecks clean, eslint is clean on the component, and both files pass `prettier --check`.

## Scope

Deliberately narrow. The same package has other gaps worth looking at separately, most obviously that the message list is not marked up as a log, so newly arrived messages are not announced at all. I would rather land this small piece and discuss that one on its own than widen the change here. Happy to follow up if it would be useful.

### DIFF
--- a/packages/jupyter-chat/src/components/writing-indicator.tsx
+++ b/packages/jupyter-chat/src/components/writing-indicator.tsx
@@ -92,6 +92,18 @@ export function WritingIndicator(
   return (
     <Box
       className={WRITERS_ELEMENT_CLASSNAME}
+      // The indicator already says something useful, "Alice is typing..." or
+      // "Jupyternaut is running `ripgrep`", but only on screen. Announcing it
+      // politely means a screen reader user learns a reply is coming without
+      // being interrupted mid-sentence.
+      //
+      // The region is the container rather than the text, so it is present in
+      // the accessibility tree before a writer appears and the change is
+      // announced. `aria-atomic` keeps the phrase together: without it a name
+      // change alone can be read out on its own, stripped of its context.
+      role="status"
+      aria-live="polite"
+      aria-atomic="true"
       sx={{
         ...props.sx,
         minHeight: '16px'
--- a/packages/jupyter-chat/src/__tests__/writing-indicator.spec.tsx
+++ b/packages/jupyter-chat/src/__tests__/writing-indicator.spec.tsx
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+import React, { act } from 'react';
+import { createRoot, Root } from 'react-dom/client';
+
+// React 18 asks test environments to declare themselves, otherwise every
+// `act` call warns.
+(
+  globalThis as unknown as { IS_REACT_ACT_ENVIRONMENT: boolean }
+).IS_REACT_ACT_ENVIRONMENT = true;
+
+import { WritingIndicator } from '../components/writing-indicator';
+import { IChatModel } from '../model';
+import { IUser } from '../types';
+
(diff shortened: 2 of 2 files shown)


dlqqq:
Welcome @opensource-joe ! Thanks for contributing this, I did not know about these attributes, appreciate the help on the a11y front.

This PR looks very safe, and I verified that each of these attributes do exactly what is claimed. I will try to include this in the pre-release today. :)

The `main` branch is in development currently prepping for the v0.25.0 release, which includes some major code changes (see [context](https://jupyter-ai.readthedocs.io/en/stable/roadmap/decouple-rtc.html)) . So if you need it sooner in an official release, please ask your AI to help backport this to the `0.24.x` branch and I can help cut a patch as well.

dlqqq (approved):
Not gonna wait for full CI, test added here is passing. Merging!

opensource-joe:
@dlqqq, I appreciate the quick turnaround; this one should be stable. I'm checking on #2035 since it's not passing CI. Trying to add contributions where it makes sense and don't want to overwhelm you. If you have specific issues, PRs, etc., you want me to view, please tag me.

dlqqq:
@opensource-joe No worries, happy to review whenever I can, it's clear you know what you are doing. :)

If want to stay in the loop on all of the ongoing initiatives in Project Jupyter (or Jupyter AI 😉), check out our community on Zulip! I'm sure you can find ways to contribute: jupyter.zulipchat.com

packages/jupyter-chat/src/components/writing-indicator.tsx
```diff
@@ -92,6 +92,18 @@ export function WritingIndicator(
   return (
     <Box
       className={WRITERS_ELEMENT_CLASSNAME}
+      // The indicator already says something useful, "Alice is typing..." or
```
brichet, line 95: We should probably clean this comment in a follow up PR, There doesn't seem to be a special use case here, it mostly documents what are the attributes.

brichet (commented):
Thanks @opensource-joe for this change, it's really nice.

PR #520, opened 2026-08-20
Backport #519 to 0.24.x: announce the writing indicator to assistive technology

opensource-joe:
Backport of #519 to `0.24.x`, as @dlqqq suggested there.

To be clear about the motive: I do not need this in a release myself. `main` is prepping v0.25.0 with the RTC decoupling behind it, so the stable line could be waiting a while, and this seemed worth having there for people on it sooner rather than later.

## The component change is identical to #519

Three attributes on the existing container, no visual change, no new dependency. Cherry-picked cleanly, no conflicts.

## The test differs in exactly one case, and that is worth flagging

`main` asserts that a custom `typingIndicator` renders, "Jupyternaut is running `ripgrep`". **That feature does not exist on `0.24.x`**: `formatWritersText` here has no branch for it, and the component never reads it. The cherry-picked test therefore failed on this branch with:

```
● announces a custom indicator, not just a generic one
    Expected substring: "Alice is running `ripgrep`"
    Received string:    "Alice is typing..."
```

I replaced that one case with multiple-writer phrasing, which this branch does support. The other four are unchanged.

## Verification on this branch, not carried over from #519

- Full package suite green: **7 suites, 57 tests**
- `build:lib` typechecks clean
- `prettier --check` clean on both files
- The two accessibility assertions fail against the parent commit, so they are not vacuous:

```
✕ is a polite live region, so a reply on its way is announced
    Expected: "status"   Received: null
✕ reads the whole phrase rather than a fragment of it
    Expected: "true"     Received: null
```

The other three pass either way by design. They are regression guards on behaviour this must not alter.

### DIFF
--- a/packages/jupyter-chat/src/components/writing-indicator.tsx
+++ b/packages/jupyter-chat/src/components/writing-indicator.tsx
@@ -76,6 +76,18 @@ export function WritingIndicator(
   return (
     <Box
       className={WRITERS_ELEMENT_CLASSNAME}
+      // The indicator already says something useful, "Alice is typing..." or
+      // "Jupyternaut is running `ripgrep`", but only on screen. Announcing it
+      // politely means a screen reader user learns a reply is coming without
+      // being interrupted mid-sentence.
+      //
+      // The region is the container rather than the text, so it is present in
+      // the accessibility tree before a writer appears and the change is
+      // announced. `aria-atomic` keeps the phrase together: without it a name
+      // change alone can be read out on its own, stripped of its context.
+      role="status"
+      aria-live="polite"
+      aria-atomic="true"
       sx={{
         ...props.sx,
         minHeight: '16px'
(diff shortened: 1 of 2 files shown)


dlqqq (approved):

packages/jupyter-chat/src/components/writing-indicator.tsx
```diff
@@ -76,6 +76,18 @@ export function WritingIndicator(
   return (
     <Box
       className={WRITERS_ELEMENT_CLASSNAME}
+      // The indicator already says something useful, "Alice is typing..." or
+      // "Jupyternaut is running `ripgrep`", but only on screen. Announcing it
```
brichet, line 80: Same comment as in https://github.com/jupyterlab/jupyter-chat/pull/519#discussion_r3841802267.

Additionally the comment is unfair as in this branch there is no way to write `Jupyternaut is running ripgrep`, only is typing is available.

PR #521, opened 2026-08-20
fix: avoid background tasks for synchronous YChat updates

### LINKED ISSUE #509: Unnecessary async background tasks in YChat
## Description

In `jupyter-ai-persona-manager`, we noticed an issue where background tasks had to be awaited during the server shutdown sequence.

https://github.com/jupyter-ai-contrib/jupyter-ai-persona-manager/blob/6adda36d188070a5b7cc510794af8db4f6510c1e/jupyter_ai_persona_manager/persona_manager.py#L627-L630

I came across this again when refactoring the Jupyter AI code. But I don't think this is necessary. These tasks are only created to run these 2 methods, neither of which are truly async.

```py
    async def create_id(self) -> str:
        """
        Creates a new ID for the document.
        """
        id = str(uuid4())
        self.set_id(id)
        return id

```

```py

    async def _set_timestamp(self, msg_idx: int, timestamp: float):
        """
        Update the timestamp of a message and reinsert it at the correct position.
        """
        with self._ydoc.transaction():
            # Remove the message from the list and modify the timestamp
            try:
                message = self._ymessages[msg_idx]
            except IndexError:
                return

            message.update({"time": timestamp, "raw_time": False})  # type:ignore[index]

            # Move the message at the correct position in the list, looking first at the end, since the message
            # should be the last one.
            # The next() function below return the index of the first message with a timestamp inferior of the
            # current one, starting from the end of the list.
            new_idx = len(self._ymessages) - next((i for i, v in enumerate(self._get_messages()[::-1]) if v.get("time", 0) < timestamp), len(self._ymessages))
            if msg_idx != new_idx:
                message = self._ymessages.pop(msg_idx)
                self._ymessages.insert(new_idx, message)
```

There are no `await` calls anywhere in `ychat.py`. I think this can be removed entirely and we can guarantee that `YChat` objects can be freed from memory normally like other objects.

uuzzrm:
Fixes #509

## Summary

- Keep `YChat.create_task()` available for genuine asynchronous extension work.
- Make the internal `create_id()` and `_set_timestamp()` updates synchronous.
- Schedule those document mutations on the next event-loop tick from YDoc observers.
- Add regression coverage for ID initialization, timestamp normalization, message observer delivery, and the absence of internal background tasks.

## Why

`create_id()` and `_set_timestamp()` do not await anything, so wrapping them in tracked background tasks adds unnecessary lifetime and scheduling overhead. Calling them directly from a YDoc observer is not safe, though: observer callbacks run inside a read-only transaction, and the write would fail. Deferring the synchronous callback by one event-loop tick preserves the transaction boundary without creating a task that retains the chat document.

## Testing

- `pytest -vv -r ap --cov` — 73 passed, 2 skipped (RTC integration tests)
- `mypy python/jupyterlab-chat` — passed
- `git diff --check` — passed

### DIFF
--- a/python/jupyterlab-chat/jupyterlab_chat/tests/test_message_observer.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/tests/test_message_observer.py
@@ -5,6 +5,7 @@
 Covers ``observe_messages``/``unobserve_messages`` on both the WebSocket model
 (RTC-free) and the collaborative ``YChat``.
 """
+import asyncio
 from pathlib import Path
 from typing import List
 
@@ -115,8 +116,8 @@ def test_ws_observer_error_is_isolated(tmp_path: Path) -> None:
 # Collaborative model (YChat)
 # --------------------------------------------------------------------------
 def _insert_ymessage(chat: YChat, message: dict) -> None:
-    # raw_time=False avoids the server-timestamp reschedule (which uses an
-    # asyncio task) so the test stays synchronous.
+    # raw_time=False avoids the server-timestamp reschedule so the test stays
+    # synchronous.
     message.setdefault("raw_time", False)
     with chat._ydoc.transaction():
         chat._ymessages.append(Map(message))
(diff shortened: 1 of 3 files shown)


dlqqq:
Ah, interesting, so there is a reason they were done async - they had to be done on the next tick of the event loop. Great find, thank you!

dlqqq:
I will hold off on merging this *just in case* there is something I don't fully understand which could break things. @brichet , can you help check if this looks good?

dlqqq (approved):

dlqqq:
Since this is just going to a pre-release, I think it is fine to merge. Nicolas will be out for a few days.

E2E tests should have caught any errors, and if not, we'll learn something from the experience. Manual testing will reveal whether there are any actual issues with this.

Thank you for fixing this! @uuzzrm

Issue #522, opened 2026-08-20
Don't free a chat room until everyone is done writing


dlqqq:
## Description

Right now, the WebSocket chat will stop a chat as soon as the last client disconnects:

https://github.com/jupyterlab/jupyter-chat/blob/9db2b7fb7fd3b4dab4c5a66a8f3ef89440599e26/python/jupyterlab-chat/jupyterlab_chat/events.py#L333-L342

dlqqq:
We should also add a 1 minute timeout so the room doesn't just die as soon as we disconnect

dlqqq:
So to be clear

- in RTC-free case : we rely on our own logic for automatically cleaning and freeing up rooms from memory

- with RTC: we rely on parent extension to free rooms, and we just respond to that

For the E2E test in the latter case, we can just simulate that by emitting the event ourselves. We should trust the RTC provider to handle memory management properly and leave that out of scope

dlqqq:
Fixed by #535

PR #525, opened 2026-08-21
Add get_path() to server-side chat models

### LINKED ISSUE #524: Add `get_path()` to server-side chat models
### Problem

There should be an abstract required `BaseChatModel.get_path()` method to get the path of the file backing the chat model. This path is the path relative to the contents root.

This can be a stopgap until #522 updates `WsChatModel` to also return a path.

### Proposed Solution

- Implement `BaseChatModel.get_path()` as required method, returning a string that resolves back to the path of the chat relative to `ContentsManager.root_dir`.
- Safely get a reference to the file ID manager, but don't assume `jupyter_server_fileid` is installed for now (until #522 is in place)
- Use FileID manager to get the path for `YChat` (since it will always be installed if RTC is installed)
- For `WsChatModel`, just return the path since it doesn't use file IDs yet.

dlqqq:
Fixes #524

## Summary
Adds a required `BaseChatModel.get_path()` that returns the chat file's path relative to `ContentsManager.root_dir`, so consumers can locate the backing file without reaching into transport-specific internals.

`WsChatModel` returns its path directly (it does not use file IDs yet). `YChat` prefers the File ID service so the path follows the file across moves and renames, and falls back to the path recorded in the shared document state.

## Technical details
`get_path()` does not assume `jupyter_server_fileid` is installed: `YChat` looks the manager up from the running server's settings only when it is present (it always is when RTC is enabled), and otherwise returns the shared-state path. This is a stopgap until #522 has `WsChatModel` track paths through file IDs too.

## Testing
Unit tests for both transports, including that `YChat` resolves through the File ID service when one is available.

### DIFF
--- a/python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
@@ -59,9 +59,6 @@ class WsChatModel(BaseChatModel):
         self._attachments: Dict[str, dict] = {}
         self._metadata: Dict[str, object] = {}
         self._message_observers: List[MessageObserverCallback] = []
-        # A random id assigned once per model instance. Not persisted to the
-        # chat file: no consumer needs it to be stable across reloads.
-        self._id = uuid.uuid4().hex
 
         # Track in-band moves: a rename via the ContentsManager updates our
         # tracked path, so subsequent saves go to the file's new location. This
@@ -97,6 +94,10 @@ class WsChatModel(BaseChatModel):
             self._metadata = content.get("metadata", {})
         except (FileNotFoundError, json.JSONDecodeError):
             self._metadata = {}
+        # A stable id lives in the chat file's metadata (same as the
+        # collaborative model). Generate one if the file has none yet; it is
+        # persisted on the next save.
+        self._metadata.setdefault("id", uuid.uuid4().hex)
         self._indexes_by_id = {m["id"]: i for i, m in enumerate(self._messages) if "id" in m}
 
     def save(self) -> None:
@@ -150,7 +151,9 @@ class WsChatModel(BaseChatModel):
     # ------------------------------------------------------------------
 
     def get_id(self) -> str:
-        return self._id
+        # The id is stored in the chat file's metadata (same as the
+        # collaborative model). Create one lazily if it does not exist yet.
+        return self._metadata.setdefault("id", uuid.uuid4().hex)  # type: ignore[return-value]
 
     def get_path(self) -> str:
         # The WebSocket model does not use file IDs; its path is tracked
--- a/python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
@@ -58,11 +58,11 @@ class WsChatModel(BaseChatModel):
         self._indexes_by_id: dict[str, int] = {}
         self._users: Dict[str, dict] = {}
         self._attachments: Dict[str, dict] = {}
-        # Stable id assigned once. Persisted in the chat file's metadata, so it
-        # survives reloads and in-band moves (the file carries it to the new
-        # path). ``load_from_file`` adopts a persisted id if the file has one.
-        self._metadata: Dict[str, object] = {"id": uuid.uuid4().hex}
+        self._metadata: Dict[str, object] = {}
         self._message_observers: List[MessageObserverCallback] = []
+        # A random id assigned once per model instance. Not persisted to the
+        # chat file: no consumer needs it to be stable across reloads.
+        self._id = uuid.uuid4().hex
 
         # Track in-band moves: a rename via the ContentsManager updates our
         # tracked path, so subsequent saves go to the file's new location. This
@@ -89,7 +89,6 @@ class WsChatModel(BaseChatModel):
 
     def load_from_file(self) -> None:
         full_path = self.root_dir / self.path
-        existing_id = self._metadata.get("id")
         try:
             with open(full_path) as f:
                 content = json.load(f)
@@ -99,10 +98,6 @@ class WsChatModel(BaseChatModel):
             self._metadata = content.get("metadata", {})
         except (FileNotFoundError, json.JSONDecodeError):
             self._metadata = {}
-        # Guarantee a stable id: adopt the file's persisted id if present,
-        # otherwise keep the id assigned at construction.
-        if not self._metadata.get("id"):
-            self._metadata["id"] = existing_id or uuid.uuid4().hex
         self._indexes_by_id = {m["id"]: i for i, m in enumerate(self._messages) if "id" in m}
 
     def save(self) -> None:
@@ -156,7 +151,7 @@ class WsChatModel(BaseChatModel):
     # ------------------------------------------------------------------
 
     def get_id(self) -> str:
-        return self._metadata["id"]  # type: ignore[return-value]
+        return self._id
 
     def get_path(self) -> str:
         # The WebSocket model does not use file IDs; its path is tracked
--- a/python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
@@ -125,6 +125,11 @@ class WsChatModel(BaseChatModel):
     def get_id(self) -> Optional[str]:
         return self._metadata.get("id")  # type: ignore[return-value]
 
+    def get_path(self) -> str:
+        # The WebSocket model does not use file IDs; its path is fixed at
+        # construction and is already relative to the ContentsManager root.
+        return self.path
+
     def get_message(self, id: str) -> Optional[Message]:
         idx = self._indexes_by_id.get(id)
         if idx is None:
--- a/python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
+++ b/python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
@@ -7,8 +7,10 @@ import time
 import uuid
 from dataclasses import asdict
 from pathlib import Path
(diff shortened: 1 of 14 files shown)


python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
```diff
@@ -122,8 +148,59 @@ def resolve_message(self, message: dict) -> dict:
...
+        event listener so it does not outlive the model."""
+        if self._event_logger is not None:
+            try:
+                self._event_logger.remove_listener(
+                    schema_id=CONTENTS_EVENT_SCHEMA_ID,
+                    listener=self._on_contents_event,
+                )
+            except Exception:  # pragma: no cover - defensive
+                pass
+            self._event_logger = None
```
dlqqq: There is no need to unbind `_event_logger`, this can just remove the listener

python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -42,6 +42,10 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._background_tasks: Set[asyncio.Task] = set()
         self.dirty = True
+        # The RTC room id ("{format}:{type}:{file_id}"), set by whoever resolves
+        # this document (e.g. the ChatManager). Used by get_path() to recover
+        # the file id. `None` when the chat was not resolved through a room.
+        self.room_id: Optional[str] = None
```
dlqqq, line 49: This is not optional, every chat gets a room ID. Make this a required argument for `YChat`.

python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -380,11 +384,14 @@ def create_id(self) -> str:
...
+    def get_id(self) -> str:
         """
-        Returns the ID of the document.
+        Returns the ID of the document, creating one if it does not exist yet.
         """
-        return self._ymetadata.get("id", None)
+        existing = self._ymetadata.get("id", None)
+        if existing:
+            return existing
+        return self.create_id()
```
dlqqq, line 391: No, this should just return `self.room_id`.

python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
```diff
@@ -122,8 +148,59 @@ def resolve_message(self, message: dict) -> dict:
...
+
+        Covers an exact file rename (``current == source``) and the rename of an
+        ancestor directory (``current`` nested under ``source``).
+        """
+        if current == source:
+            return dest
+        prefix = source + "/"
+        if current.startswith(prefix):
+            return dest + "/" + current[len(prefix):]
+        return None
```
dlqqq: 1. This is good, but no need for static method, just create a separate `_on_path_change()` method. The `on_contents_event()` just only calls `_on_path_change()` if it affects a parent dir or the file itself.

2. Use `os.path.join()` here. Use other stdlib like `pathlib` or `os` as needed to reduce necessary code.

python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -488,7 +541,10 @@ def _initialize(self, event: MapEvent) -> None:
         """
         if self.dirty:
             return
-        if (self.get_id() is None):
+        # Read the raw metadata rather than get_id(): get_id() lazily creates an
+        # id, and this observer runs inside a read-only transaction where writes
+        # are forbidden. The (deferred) create_id below performs the write.
+        if self._ymetadata.get("id", None) is None:
```
dlqqq, line 544: If we were not storing id inside metadata on 0.24.0, then remove all occurrences of this!

python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -488,7 +541,10 @@ def _initialize(self, event: MapEvent) -> None:
...
             self._schedule_callback(self.create_id)
```
dlqqq, line 545: Again why is this needed? ID for `YChat` is just the room ID.

python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
```diff
@@ -122,8 +148,59 @@ def resolve_message(self, message: dict) -> dict:
...
-        return self._metadata.get("id")  # type: ignore[return-value]
+    def get_id(self) -> str:
+        return self._id
+
+    def get_path(self) -> str:
+        # The WebSocket model does not use file IDs; its path is tracked
+        # in-process and kept current on in-band moves (see _on_contents_event).
+        return self.path
+
+    async def _on_contents_event(self, logger, schema_id: str, data: dict) -> None:
```
dlqqq, line 163: Why is this async?

python/jupyterlab-chat/jupyterlab_chat/ychat.py
```diff
@@ -42,6 +42,10 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._background_tasks: Set[asyncio.Task] = set()
```
dlqqq: Let's remove all references to `_background_tasks` and `create_task` if they are no longer used . I believe this is the case after #521.

python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
dlqqq: Done in 6b11104 — `dispose()` now just removes the listener (dropped `self._event_logger = None`). I also registered the ContentsManager schema in the WS `ChatManager` unit tests: the previous `try/except` was silently swallowing a `KeyError` from `jupyter_events.remove_listener` when the schema wasn't registered on the (bare) test logger, so the listener was never actually removed there. With the schema registered (as a real server does), removal now succeeds and the guard is unnecessary.

python/jupyterlab-chat/jupyterlab_chat/ychat.py
dlqqq: This one is blocked by how the document is constructed. jupyter-collaboration builds the shared doc as `YDOCS.get(file_type)(self.ydoc, self.awareness)` in `DocumentRoom.__init__` (rooms.py) — it never passes a room id. Making `room_id` a required constructor argument would break RTC document creation (same for JSD).

Right now the `ChatManager` sets `room_id` immediately after `get_document(room_id=...)`. To make it a required constructor arg we would need an upstream change so the provider passes the room id into the YDoc factory. How would you like to proceed?

python/jupyterlab-chat/jupyterlab_chat/ychat.py
dlqqq: Depends on the point above: `room_id` is not passed at construction (jupyter-collaboration builds `YChat(ydoc, awareness)`), so it is `None` on a freshly constructed doc until the `ChatManager` sets it. Returning `self.room_id` from `get_id()` (typed `-> str`) would then return `None` for an un-resolved doc. I'm happy to switch `get_id()` to return `self.room_id` once we've settled how the room id reaches the doc (the comment above).

python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
dlqqq: Done in 6b11104 — replaced the static `_relocated_path` with an instance `_on_path_change()`. `_on_contents_event()` now calls it only when the rename affects the file itself or an ancestor directory, and uses `os.path.join` / `os.path.relpath`.

python/jupyterlab-chat/jupyterlab_chat/ychat.py
dlqqq: Checked: v0.24.1 **does** store `id` in metadata. `git show v0.24.1:python/jupyterlab-chat/jupyterlab_chat/ychat.py` has `create_id`/`set_id`, `_initialize` scheduling `create_id()`, and `get_id()` returning `self._ymetadata.get("id")`. So the condition here isn't met, and I did not remove all occurrences — dropping the metadata `id` changes the persisted `.chat` file format from 0.24.x. Do you want to drop it anyway for 0.25?

python/jupyterlab-chat/jupyterlab_chat/ychat.py
dlqqq: Same as the note on the line above: the metadata `id` (and this `_initialize` creation) has been part of the file format since at least v0.24.1, so removing it is a format change rather than dead-code cleanup. If YChat's id becomes the room id, I agree this can go — holding until we settle that / the 0.24.0 question.

python/jupyterlab-chat/jupyterlab_chat/websocket_model.py
dlqqq: It has to be `async`: `jupyter_events` dispatches listeners via `asyncio.create_task(listener(...))` in `EventLogger.emit`, so a listener must be a coroutine. The actual path-update logic is now synchronous in `_on_path_change()` (per your other comment); `_on_contents_event` is just the required async entry point.

python/jupyterlab-chat/jupyterlab_chat/ychat.py
dlqqq: Done in 6b11104 — removed `create_task` and `_background_tasks`. Confirmed they were unused after #521 (only the two `*_without_a_background_task` test assertions referenced them; updated those tests).

dlqqq:
OK, this took way longer than expected, but now, the new methods available on `WsChatModel` and `YChat`:

```
get_path() # returns the path of the chat file, responding to out-of-band moves
get_id()   # returns a unique ID for this chat, reads and sets chat.metadata.id (not the same as room ID)
```

The ID is also now available on the frontend for consumers.

PR #526, opened 2026-08-22
Make `chat_id` a reliable identifier for every chat

dlqqq:
## Summary
Makes the chat's stable id (`chat.get_id()`) the transport-neutral source of truth for chat lifecycle events, and guarantees the frontend `model.id` matches the backend.

## Changes
- **Split `events.py`**: the schema/`ChatEvent` dataclass stay in `events.py`; the `ChatManager` moves to a new `chat_manager.py` (the old placement was confusing). Import path has been changed (`from jupyterlab_chat.chat_manager import ChatManager`).

- **`chat_id`, not `room_id`, in the event**: `CHAT_ROOM_EVENT_SCHEMA`/`ChatEvent` now carry `chat_id` (`chat.get_id()`) and drop `room_id`. `room_id` remains internal to the ChatManager's RTC resolution and to `YChat` (recovering the file id for `get_path()`); it is never emitted. `chat_id` is captured before a model is freed, so `closed`/`deleted` events still carry it, and the RTC `opened` event reads it only after `get_document()` has loaded content (so it never mints a premature id that could conflict with disk).

- **Frontend `ready` carries the id**: `IChatModel.ready` is now `Promise<string>` resolving with the chat id, so `const id = await model.ready` is the guaranteed, backend-matching id. The WebSocket path no longer mints a local `UUID` fallback: the id is exactly the server's `chat.get_id()` delivered in the connection frame. Under RTC the id continues to come from the synchronized shared-document metadata (same value the backend persists).

### DIFF
--- a/docs/jupyter-chat-example/src/index.ts
+++ b/docs/jupyter-chat-example/src/index.ts
@@ -52,7 +52,8 @@ class ChatContext extends AbstractChatContext {
 class MyChatModel extends AbstractChatModel {
   constructor(options: IChatModel.IOptions) {
     super(options);
-    this.setReady();
+    // This example model has no backend; give it a stable client-side id.
+    this.setReady(UUID.uuid4());
   }
 
   sendMessage(newMessage: INewMessage): string | null {
--- a/packages/jupyter-chat/src/model.ts
+++ b/packages/jupyter-chat/src/model.ts
@@ -1011,9 +1011,11 @@ export namespace IChatModel {
  */
 export interface IChatContext {
   /**
-   * The unique id of the chat, when known.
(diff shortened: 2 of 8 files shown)


dlqqq:
The existing `ChatManager` code was super messy and had like 3 different ways of looking up a chat. It now just maintains a single source of truth `chat_id` -> `BaseChatModel`. 

Every chat event also now includes the `chat_id`.

dlqqq:
Note that this does include one breaking change, chat models in the frontend have to call `this.setReady(<uuid>)` to provide a way for frontend consumers to get guaranteed access to the chat ID. To be safe, I'm labeling this an API change.

PR #529, opened 2026-08-24
Jupyterlite compatibility

brichet:
This PR makes `jupyterlab-chat` extension compatible with Jupyterlite:
- if RTC and websocket is not available, fallback to write in the shared model (as if RTC was available), which will trigger a signal updating the model with the new message.
- in this case (RTC and WS unavailable), the message `raw_time` is set to false, because there is no server as source of truth

It also updates the lite deployment in docs, to use `jupyterlab-chat`  extension directly, instead of deploying a lite specific extension  (`jupyter-chat-example`).


brichet:
The Jupyterlite deployment seems to work as expected https://jupyter-chat--529.org.readthedocs.build/en/529/lite/lab/index.html

packages/jupyterlab-chat/src/model.ts
```diff
@@ -326,7 +326,17 @@ export class LabChatModel
...
+          console.warn(
+            'WS chat connection failed, falling back to shared model',
+            e
+          );
+          this._wsHandler?.dispose();
+          this._wsHandler = null;
+          if (!this._sharedModel.id) {
+            this._sharedModel.id = UUID.uuid4();
+          }
+        });
```
dlqqq, line 350: This requires a failing WS connection to be attempted first, which is a little weird (although harmless).

Is there a more reliable way to detect whether this labextension is in a Jupyterlite environment or not? With that, we can avoid starting `this._wsHandler` at all in JupyterLite.

dlqqq (approved):
@brichet Thank you! Sorry I broke JupyterLite, but I'm glad the fix was fairly straightforward. Just one small comment, but it's not blocking.

Also, I see `docs/jupyter-chat-example` isn't needed anymore since this now works with JupyterLite, which is a neat side-effect. 👍

dlqqq:
Sorry for introducing conflicts with this by accident after my changes to the WS API (#530). Happy to help fix this, just need some details on how to verify Jupyterlite compatibility.

Is there a way we could add a separate E2E test to verify things stay Jupyterlite compatible?

packages/jupyterlab-chat/src/model.ts
brichet: I don't know if there has been some changes since 2023 about that, but at the time is was intentionally not possible: https://github.com/jupyterlite/litegitpuller/issues/5#issuecomment-1705342447

brichet:
> Sorry for introducing conflicts with this by accident after my changes to the WS API (https://github.com/jupyterlab/jupyter-chat/pull/530)

The changes in #530 are really similar about handling a connection error. The main difference is that the model ends up removing the WS handler in the current PR, to allow updating the shared model when WS is not available.
I'll rebase it.

packages/jupyterlab-chat/src/model.ts
brichet: > This requires a failing WS connection to be attempted first, which is a little weird (although harmless).

I updated it to handle only the case where server is unreachable (error 1006), instead of any error during connection. If there a error in path while using the websocket handler, it keep gracefully close the chat as not available, instead of creating a server-less chat.

brichet:
Merging it as it has already be approved, the tests pass and the document deployment works fine.
In follow up PR, we'll need to:
- sync and save file in jupyterlite, current the chat are transient in jupyterlite, the save methods are only on server (collaboration or web socket)
- improve UI tests to handle jupyterlite case

PR #530, opened 2026-08-25
Clean up chat WebSocket API

dlqqq:
## Motivation

The RTC-free chat WebSocket endpoint was named `api/jupyter-chat/ws`. The `jupyter-` prefix is redundant for an endpoint already served by a Jupyter server, and inconsistent with the rest of the API. The handshake also carried the chat path and the connecting user's identity as query arguments — more surface area than this transport needs.

While cleaning it up, two related issues surfaced: a chat that failed to open left a loading spinner running forever in the side panel, and the server never checked that the requested path stayed within its root.

## Summary

The endpoint is now `api/chat/ws/<path>`, with the chat file path carried as a single URL path segment instead of a query argument.

Identity is now owned by the server. The RTC-free transport is single-user, so instead of the client asserting who it is, the server registers the connection's authenticated user and reports that identity back on the connection frame; the client adopts it. Client and server therefore always agree on who the user is, which is what drives sender attribution and excluding yourself from mention suggestions. A chat only becomes ready once it has received both its id and this identity.

A chat that cannot be opened now fails gracefully instead of hanging. When the server rejects the connection, a chat opened in the side panel is closed (the panel itself stays open), a chat opened in the main area disposes its widget, and an error notification is shown — no more spinner spinning forever.

The server also validates that the path resolves inside the ContentsManager root and closes the connection otherwise. This closes a path-traversal hole — a client could previously pass something like `../../another-user/private.chat` — though the risk was low since the endpoint is authenticated.

**This is a breaking change to the WebSocket wire API** (endpoint URL and handshake). That is fine at this alpha stage; the RTC-free transport has no stable released consumers yet.

## Technical details

- The path segment is percent-encoded on the client (`encodeURIComponent`) and read back from the route on the server (which url-unescapes it), so there is no custom escaping on either side.
- The client no longer sends its identity — neither the connect query argument nor the per-message `user`. The server stamps every sender from `current_user` and returns that identity in the connection frame for the client to adopt.

## Testing

Added a unit test for the path-safety check and an end-to-end test that opens a chat at a rejected path and asserts the side panel closes it and shows the notification; it was verified failing before the fix (the spinner hangs) and passing after. The integration server now reports a deterministic identity so the identity-dependent tests and screenshots are reproducible.

### DIFF
--- a/packages/jupyter-chat/src/widgets/multichat-panel.tsx
+++ b/packages/jupyter-chat/src/widgets/multichat-panel.tsx
@@ -8,7 +8,11 @@
  * Originally adapted from jupyterlab-chat's ChatPanel
  */
 
-import { InputDialog, ToolbarRegistry } from '@jupyterlab/apputils';
+import {
+  InputDialog,
+  Notification,
+  ToolbarRegistry
+} from '@jupyterlab/apputils';
 import { IObservableList } from '@jupyterlab/observables';
 import { nullTranslator, TranslationBundle } from '@jupyterlab/translation';
 import {
@@ -554,9 +558,23 @@ class SidePanelWidget extends ReactivePanelWithToolbar implements IChatPanel {
     // Add spinner while loading
     const spinner = new Spinner();
     this.addWidget(spinner);
-    this._chatWidget.model.ready.then(() => {
-      spinner.dispose();
-    });
+    this._chatWidget.model.ready
+      .then(() => {
+        spinner.dispose();
+      })
+      .catch(() => {
+        // The chat could not be opened (e.g. the server closed the WebSocket
+        // because the path was invalid). Dispose the spinner and close this
+        // chat so no loading spinner is left hanging; the side panel stays open.
+        spinner.dispose();
+        Notification.error(
+          trans.__(
+            "Unable to open chat at given path: '%1'.",
+            this._chatWidget.model.name
+          )
+        );
+        options.onClose(this._displayName);
+      });
 
     // Add the chat widget
     this.addWidget(this._chatWidget);
--- a/packages/jupyterlab-chat/src/websocket-handler.ts
+++ b/packages/jupyterlab-chat/src/websocket-handler.ts
@@ -95,6 +95,15 @@ export class WebSocketHandler {
     return this._chatId;
   }
 
+  /**
+   * The identity the server registered for this connection, received on the
+   * connection message. Available once `ready` has resolved; `undefined`
+   * against older servers that do not send it.
+   */
+  get connectedUser(): IUser | undefined {
+    return this._connectedUser;
+  }
+
   setPath(path: string): void {
     this._path = path;
   }
@@ -157,6 +166,7 @@ export class WebSocketHandler {
   private _handleMessage(data: any): void {
     if (data.type === 'connection') {
       this._chatId = data.id as string | undefined;
+      this._connectedUser = (data.user as IUser) ?? undefined;
       this._usersMap = (data.users as Record<string, IUser>) ?? {};
       this._usersChanged.emit(this._usersMap);
       for (const msg of (data.messages as any[]) ?? []) {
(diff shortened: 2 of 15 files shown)


python/jupyterlab-chat/jupyterlab_chat/websocket_handler.py
```diff
@@ -14,14 +15,36 @@
...
     One instance per connected client; all clients connected to the same
     .chat file share a WsChatModel; the registry lives in settings["chats_by_id"].
     """
-    _path: str
+    # A real default (not just an annotation) so ``self._path`` resolves even
+    # when the connection closes before ``open()`` runs -- e.g. a connection
+    # that never sends a decodable chat path, causing ``open()`` to
+    # ``self.close()`` and return early. Both ``on_message`` and ``on_close``
+    # treat a falsy ``_path`` as "never opened" and return.
+    _path: str = ""
```
dlqqq: Doesn't this technically define this to be a class attribute? I would just set `self._path` somewhere else.

dlqqq: Shouldn't we just close the WebSocket with a reasonable error code in the case where `path` was not specified or is unparseable?

The frontend should handle this gracefully too, I think right now it just hangs forever with an orange spinner. Add a failing E2E test first and verify that if the WebSocket gets closed, it gracefully returns to the original state before the user sent the request, and emits an error notification like "Unable to open chat at given path: '<path>'." To reproduce this error, try requesting an obviously invalid path.

packages/jupyterlab-chat/src/websocket-handler.ts
```diff
@@ -9,19 +9,29 @@ import { ServerConnection } from '@jupyterlab/services';
...
+  const bytes = new TextEncoder().encode(path);
+  let binary = '';
+  for (const byte of bytes) {
+    binary += String.fromCharCode(byte);
+  }
+  return btoa(binary)
+    .replace(/\+/g, '-')
+    .replace(/\//g, '_')
+    .replace(/=+$/, '');
+}
```
dlqqq: Why is this done so manually? Is there no standard library for this?

ui-tests/tests/user-mention.spec.ts
```diff
@@ -160,9 +160,11 @@ test.describe('#bot-mention', () => {
...
   });
 
-  test.use({
-    mockUser: USER
-  });
+  // Intentionally no `mockUser`: the RTC-free WebSocket registers the
+  // authenticated server user as the connected user, so the frontend must use
+  // that same server identity (not a mocked one) for it to be recognised as
+  // "self" and excluded from mention suggestions -- as it is in a real
+  // single-user deployment.
```
dlqqq: This can be dropped.

packages/jupyter-chat/src/widgets/multichat-panel.tsx
```diff
@@ -554,9 +558,23 @@ class SidePanelWidget extends ReactivePanelWithToolbar implements IChatPanel {
...
+        // because the path was invalid). Dispose the spinner and close this
+        // chat so no loading spinner is left hanging; the side panel stays open.
+        spinner.dispose();
+        Notification.error(
+          trans.__(
+            "Unable to open chat at given path: '%1'.",
+            this._chatWidget.model.name
+          )
+        );
+        options.onClose(this._displayName);
```
dlqqq, line 576: This fixes the orange spinner appearing forever in the side panel in case the chat cannot be opened for whatever reason. Let's update the PR description.

python/jupyterlab-chat/jupyterlab_chat/websocket_handler.py
```diff
@@ -45,37 +74,14 @@ async def get(self, *args, **kwargs):
...
-        if path is None:
-            self.close(1008, "Missing 'path' query parameter")
+        # tornado url-unescapes the captured route segment, so ``args[0]`` is the
+        # decoded chat path (with slashes restored).
+        path = args[0] if args else ""
+        if not is_safe_chat_path(path, self._root_dir):
+            # 1008 (policy violation): the path is missing, unparseable, or would
+            # escape the server root. The frontend surfaces this to the user.
+            self.close(1008, "Invalid chat path")
             return
```
dlqqq, line 86: This fixes a path traversal vulnerability (before, clients could pass `../../../home/someone-else/...`). Not really a concern since the WS is authenticated, but still good to note that we now validate that the given initial path is under the ContentsManager root dir. Let's update the PR description as well.

PR #531, opened 2026-08-27
Push user and metadata updates to WS chat clients

dlqqq:
## Motivation

In RTC-free (WebSocket) chats, a user added to a chat after clients have already connected never shows up for those clients. This bites AI personas in particular: a persona registers itself with `set_user()` once it is initialized, which happens after the web client connects, so the persona's name and avatar never reach the client. Chat metadata set after connect has the same problem.

The underlying reason is that the WS protocol was untyped dicts on both ends, and the only place that ever broadcast the users map was the connection handshake. `set_user`/`set_metadata` mutated server state silently.

## Summary

`set_user` and `set_metadata` now push an update to every connected client, so personas (and any later-added user or metadata) appear live.

The `/api/chat/ws` protocol is now a typed, discriminated schema instead of free-form dicts. Every frame carries `type` (`client` or `server`, the direction) and `action` (the specific message). The backend defines it in a dedicated `ws_messages` module and the frontend mirrors it in `ws-messages.ts`; both the handler and the model construct and parse frames through the schema, and malformed client frames are rejected rather than half-handled.

This changes the wire format, so the server and the labextension must be upgraded together.

## Technical details

- The schema lives in its own module rather than inside `websocket_handler.py`, because the handler imports the model and both need the schema — defining it in the handler would be circular.
- `open()` now calls `set_user` before registering its own handler, so already-connected clients get the new user via the normal broadcast and the connecting client still gets the full users map in its connection frame (the ad-hoc users broadcast in `open()` is gone).
- Frontend gains a `metadataChanged` signal on the WS handler and a `setMetadata` on the shared model to apply metadata updates.

## Testing

Added Python tests that assert `set_user`/`set_metadata` broadcast to connected clients (the reported bug, red before the fix), plus schema round-trip/parse tests. Existing writing-status tests updated to the new frame shape. Frontend types checked via the build; jest suite green.

### DIFF
--- a/packages/jupyterlab-chat/src/model.ts
+++ b/packages/jupyterlab-chat/src/model.ts
@@ -234,6 +234,7 @@ export class LabChatModel
       });
       this._wsHandler.messageReceived.connect(this._onWsMessage, this);
       this._wsHandler.usersChanged.connect(this._onWsUsersChanged, this);
+      this._wsHandler.metadataChanged.connect(this._onWsMetadata, this);
       this._wsHandler.writingChanged.connect(this._onWsWriting, this);
     }
   }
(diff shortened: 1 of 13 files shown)


dlqqq:
This fixes a bug in the WS chat, which Jupyter AI is the only consumer of currently as v0.25.0 is still in pre-release stage. None of the changes substantially touch the shared frontend part that JupyterLite relies on, so this should not break Jupyterlite in any way.

Merging.

PR #533, opened 2026-08-27
Load and save content from/to file in jupyterlite

brichet:
This PR allows to load and save content from/to file in a jupyterlite environment (no RTC, no web server).

Follow up #529 

It can be tested from the readthedoc deployment https://jupyter-chat--533.org.readthedocs.build/en/533/lite/lab/index.html

### CHANGES

If RTC and web socket are not available, the content is loaded from the frontend contents manager, and saved when the shared model changed.
It also fix the `YChat` set source, that wasn't used before the PR.

### DIFF
--- a/packages/jupyterlab-chat-extension/src/index.ts
+++ b/packages/jupyterlab-chat-extension/src/index.ts
@@ -1348,7 +1348,10 @@ const chatPanel: JupyterFrontEndPlugin<MultiChatPanel> = {
       );
       const names: { [name: string]: string } = {};
       for (const file of dirContents.content) {
-        if (file.type === 'file' && file.name.endsWith(chatFileExtension)) {
+        if (
+          (file.type === 'file' || file.type === chatFileType.contentType) &&
+          file.name.endsWith(chatFileExtension)
+        ) {
           const nameWithoutExt = file.name.replace(chatFileExtension, '');
           names[nameWithoutExt] = file.path;
         }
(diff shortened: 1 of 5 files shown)


dlqqq (approved):
Rebased conflicts with #532, I just made a change to call `this.setReady` in the new code.

I do think this is an example of why we may want to split the frontend models into `LabChatModel` and `LiteChatModel`, and pick the right model instead of relying on "WS fails => we are in JupyterLite". Right now this class is trying to support both and it's hard to tell which methods are used in JupyterLab v.s. JupyterLite.

Alternatively, to avoid adding more edge cases to support, I think we should consider dropping `YChat`. My main argument for this is that there are no conflicts to resolve in a chat - each client can only edit their own messages. The only thing we rely on RTC for is the automatic broadcasting and synchronization, which I don't think really warrants maintaining a CRDT that is only used in one circumstance or another.

dlqqq:
https://github.com/user-attachments/assets/e93aa202-3f23-465f-bb83-a61a3328f2d8

@brichet Tried the loading / saving behavior on JupyterLite, works well. But, is it expected for the messages to "pop in"? I thought it should show an orange spinner until `ready` is set and messages have arrived.

Since this seems like it may be a separate bug, I will merge this now and let you address this in another PR if needed. Not sure what the original JupyterLite experience was.

brichet:
Thanks for the review and test @dlqqq 

> But, is it expected for the messages to "pop in"? I thought it should show an orange spinner until ready is set and messages have arrived.

I haven't noticed it, I'll take a look at it.

PR #534, opened 2026-08-27
Drop the re-broadcast required for writing status in WS chat

dlqqq:
## Motivation

A server-pushed writing status (e.g. an AI persona's "is typing" indicator over the RTC-free WebSocket) was only kept alive on the client for 3 seconds. To keep the indicator visible, the sender had to re-broadcast the status on a timer; a single call would silently disappear after 3s. This was a footgun for anyone driving the indicator from the server, and it did not match how every other WebSocket update (messages, users, metadata) behaves — those persist once sent.

## Summary

A writing status now persists until it is explicitly cleared. Calling `broadcast_writing_status(user, status)` sends one frame that shows the indicator and keeps it visible; sending a stop frame (`status=None`) removes it. No re-broadcasting, no background timer.

The lifecycle is now fully explicit start/stop, controlled by the sender.

## Technical details

The auto-expiry lived entirely on the frontend: `_onWsWriting` armed a 3s timeout that dropped the writer unless refreshed. That path and the underlying per-writer timer machinery are removed, and the now-unused `timeout` parameter is dropped from `IChatModel.setWritingStatus` (a small API change).

The server model already sent a single frame like any other update, so it is unchanged. The local user's own typing debounce (RTC awareness) is unrelated and left as-is.

## Testing

Updated the writers unit test to assert a writer persists past the old timeout and is only removed on an explicit clear.

### DIFF
--- a/packages/jupyter-chat/src/__tests__/writers.spec.ts
+++ b/packages/jupyter-chat/src/__tests__/writers.spec.ts
@@ -47,20 +47,18 @@ describe('writers state', () => {
     expect(changes.length).toBe(count);
   });
 
-  it('auto-clears a writer after the timeout unless refreshed', () => {
+  it('keeps a writer until it is explicitly cleared (no auto-expiry)', () => {
     jest.useFakeTimers();
     try {
-      model.setWritingStatus(userA, undefined, 2000);
+      model.setWritingStatus(userA);
       expect(model.writers).toHaveLength(1);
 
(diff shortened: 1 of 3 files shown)


dlqqq:
The original motivation for this was to support showing writing status on clients and make sure those get automatically cleaned up on disconnect. But since `WsChatModel` is only used in the RTC-free case, and all web clients share the same identity (as of #531), this is no longer necessary. Merging.

PR #536, opened 2026-08-27
Rename Playwright report artifacts

brichet:
Rename the report artifacts of UI tests, with e.g. `integration-jupyterlab-ge4.6-collab` instead of `integration-jupyterlab-0`.


dlqqq (approved):

dlqqq:
Noticed this too, thanks for fixing it!

PR #537, opened 2026-08-27
[0.24.x] Backport PR #532: Set ready after load complete

brichet:
> Wait for the file to be synced before setting the ready state to `true` in the model.
> 
> Before this PR, `setReady()` was called as soon as the chat ID was set. To handle correctly the unread state, we need to wait for the ID to be populated, so we were waiting for the model to be `ready`, because the value is stored (local storage of the browser) according to the chat ID..
> If an extension is listening for new messages only, it should be able to know when the file is synchronized before listening for message updated signal.
> 
> This PR fixes it by:
> - waiting for the ID to be set before adding the messages, to keep the expected behavior on unread messages. If messages are loaded before, they are kept in a temporary list.
> - when the widget context (document widget) or content provider (side panel chat) is ready, the `markDocumentSynced()` is called (as before), the messages are inserted and then the chat is set as ready

_commented in #532_


brichet:
Merging it, successfully tested with https://github.com/jupyterlite/ai/pull/363: previous messages are not responded anymore by the agent.

PR #541, opened 2026-09-10
Make the chat input placeholder configurable

bsundaram1:
## Summary

`ChatInput` hardcodes `trans.__('Type a chat message, @ to mention...')` with no prop, config key, or settings entry. jupyter-ai 2.x rendered ``placeholder={`Ask ${props.personaName}`}`` before the chat UI moved into this package, so downstreams that brand the chat around a single assistant lost that and are left overwriting the DOM node to get it back.

Adds `inputPlaceholder` to `IConfig`, settable from the `Chat` settings entry or programmatically through `IChatModel.config`. An empty value (the schema default) means "use the built-in placeholder".

Named `inputPlaceholder` rather than `placeholder` since this repo already uses "placeholder" for the empty-chat-body widget (`IChatBodyPlaceholderFactory`).


brichet:
Thanks @bsundaram1 for working on this.

The change looks good and it's nice to be able to update the input placeholder.
I wonder if it's actually up to the extension that uses the chat, rather than the user with settings, to update it.

One option could be to avoid the user setting, and instead let extension update it with a plugin, using `IWidgetConfig`.

```ts
const inputPlaceholder: JupyterFrontEndPlugin<void> = {
  id: 'my-extension/input-placeholder',
  description: 'Update the input placeholder',
  autoStart: true,
  requires: [IWidgetConfig],
  activate: (app: JupyterFrontEnd, widgetConfig: IWidgetConfig): void => {
    widgetConfig.config = {
      ...widgetConfig.config,
      inputPlaceholder: 'My new placeholder'
    };
  }
};
```

bsundaram1:
> Thanks @bsundaram1 for working on this.
> 
> The change looks good and it's nice to be able to update the input placeholder. I wonder if it's actually up to the extension that uses the chat, rather than the user with settings, to update it.
> 
> One option could be to avoid the user setting, and instead let extension update it with a plugin, using `IWidgetConfig`.
> 
> ```ts
> const inputPlaceholder: JupyterFrontEndPlugin<void> = {
>   id: 'my-extension/input-placeholder',
>   description: 'Update the input placeholder',
>   autoStart: true,
>   requires: [IWidgetConfig],
>   activate: (app: JupyterFrontEnd, widgetConfig: IWidgetConfig): void => {
>     widgetConfig.config = {
>       ...widgetConfig.config,
>       inputPlaceholder: 'My new placeholder'
>     };
>   }
> };
> ```

Thanks for the review @brichet - I think your suggestion is a better fit.

bsundaram1:
The failing CI seems to be related to tornado's new `StaticFileHandler` symlink restriction (6.5.9+). CI installs the extension via `jupyter labextension develop`, so the bundle is now returning 403:

```
403 GET /lab/extensions/jupyterlab-chat-extension/static/remoteEntry….js
    … is not in root static directory
```

Wondering if this is related to jupyterlab/jupyterlab#19767.

packages/jupyter-chat/src/components/input/chat-input.tsx
```diff
@@ -40,6 +40,9 @@ export function ChatInput(props: ChatInput.IProps): JSX.Element {
   const [sendWithShiftEnter, setSendWithShiftEnter] = useState<boolean>(
     model.config.sendWithShiftEnter ?? false
   );
+  const [placeholder, setPlaceholder] = useState<string>(
+    model.config.inputPlaceholder ?? ''
+  );
```
brichet: It should be replaced with the default only if undefined, not empty string (empty string is a valid placeholder IMO).

```suggestion
  const [placeholder, setPlaceholder] = useState<string | undefined>(
    model.config.inputPlaceholder
  );
```

packages/jupyter-chat/src/components/input/chat-input.tsx
```diff
@@ -110,6 +114,11 @@ export function ChatInput(props: ChatInput.IProps): JSX.Element {
 
   const inputExists = !!input.trim();
 
+  // An empty configured placeholder means 'use the default one'. It is resolved
+  // here rather than in the settings schema, whose defaults cannot be translated.
+  const inputPlaceholder =
+    placeholder || trans.__('Type a chat message, @ to mention...');
```
brichet: Use default only if undefined.

```suggestion
    placeholder ?? trans.__('Type a chat message, @ to mention...');
```

docs/source/developers/developing_extensions/using-chat-extension.md
```diff
@@ -25,6 +25,32 @@ This token is composed of:
 - `tracker`, a widget tracker that allows to track all the opened chats, and to
   retrieve the current one.
 
+### IWidgetConfig
```
brichet, line 28: Thanks for the doc :+1:

brichet (commented):
Thanks @bsundaram1, it looks good.

I just have suggestions to allow empty string as placeholder.

brichet:
Test failure is expected, it set `inputPlaceholder` to `''` and expect the placeholder to be the default, which is not the case anymore.

bsundaram1:
> Test failure is expected, it set `inputPlaceholder` to `''` and expect the placeholder to be the default, which is not the case anymore.

Ah forgot about the tests 😅

brichet (approved):
Thanls @bsundaram1, LGTM

bsundaram1:
@brichet Thanks for merging! Could you please help make this available on conda through the feedstock?

Issue #166, opened 2025-02-14
Epic: Chat Commands framework v1.0.0


dlqqq:
## Description

This is an **epic** issue that contains sub-issues. Each sub-issue describes a follow-up item in the Chat Commands framework, which should be resolved before the v1.0.0 release.

The initial implementation of the Chat Commands framework is in PR #161.

PR #542, opened 2026-09-11
Fix #176: Do not show commands menu when user is typing in a code block

### LINKED ISSUE #176: Subissue: Do not show commands menu when user is typing literally
### Problem

We should not always trigger the commands menu based on the current input. For example, the user may be editing a Markdown code block.

### Proposed Solution

1. Explore if there's any way of showing code blocks in the chat input, kind of like how Discord & Slack handle it. Right now we show all of the backticks literally.
2. Do not open commands menu when inside of a code block.

### Additional context

May get de-prioritized, depending on how challenging this is.

Yash990-bit:
Resolves #176
Part of #166

### Problem
When typing or editing inside a Markdown code block (either inline with single backticks or multi-line with triple backticks), the chat commands autocomplete menu was erroneously triggered.

### Solution
- Added `isCursorInsideCodeBlock(input, cursorIndex)` in `use-chat-commands.tsx` to detect whether the user's cursor is currently enclosed within backtick spans (handling both inline and multi-line code blocks as well as unclosed code blocks while typing).
- Added an early return from `getCommands` when `isCursorInsideCodeBlock` evaluates to `true`, preventing the commands menu from opening.
- Added comprehensive unit tests in `use-chat-commands.spec.ts`.


Yash990-bit:
Hi @brichet , could you please take a look at this PR when you get a chance?

I’ve implemented the fix for #176 and added unit tests covering inline code blocks, multi-line code blocks, and unclosed code blocks. All checks are currently passing.

Thanks!

Yash990-bit:
Hi @brichet, just following up on this PR. The fix and tests are ready, and all checks are passing. Whenever you get a chance, I’d appreciate a review. Thanks!

dlqqq:
Problem with this is that it can be expensive to re-check the entire input on every keystroke. What we probably need is some sort of rich editing capability inside the chat input itself - i.e. typing triple backticks creates a Markdown block inside the chat instead of literally just showing ```.

dlqqq:
Would love to see someone explore that idea in another PR first. For example there's some property that lets you determine whether the current cursor is inside a code block or not. That will allow this fix to be made much more performant!

Yash990-bit:
Hi @dlqqq, thanks for the review and feedback!

You make a great point about the long-term vision — having a rich-text / markdown-aware editor in the chat input (like Slack or Discord, or backed by CodeMirror) would indeed be an awesome enhancement and provide structural AST/cursor node awareness directly.

Regarding the current implementation's performance:
1. I've added a fast-path guard (`!input.includes('`')` and checking `!currentWord?.length` first). For the vast majority of typing when no backticks exist, this bails out in sub-microsecond time with zero string iteration.
2. Even when backticks are present, the single-pass linear loop takes < 0.03ms for typical chat lengths (well below a single animation frame).

Would it make sense to merge this lightweight fix to resolve #176 for now, and open a separate issue/PR to explore the rich-text / markdown input component? I'd be happy to help explore or contribute to that as well!

